PR #11, opened 2012-04-04
Issue 9 - Make the AST position aware, use that for TypeNotFoundException, and pretty print exceptions in general

sodabrew:
In these changesets, I implement TypeNotFoundException reporting with the bad type's position in the file.

This is accomplished by:

*  Switching the thrift file input from a slurp-to-String to a StreamReader. This reader annotates the input with position information
*  Adding a Positional mixing to the top level Node class of the AST.
*  Calling the Parser positioned() combinator in useful places in the parser.
*  Subclassing several exceptions from a new ExceptionAt which has a Position instance var.

Caveats:

*  Unit tests not yet updated for the change from String to StreamReader.
*  Exceptions are mostly suppressed, so weird ones might be more confusing now.
*  I've been testing these based off the sbt11 branch because I cannot build locally with sbt 0.7, so these changesets have been cherry-picked to a new branch off master (though no errors or editing needed).

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,3 +163,4 @@ included file `LICENSE`.
 - Robey Pointer
 - Ian Ownbey
 - Jeremy Cloud
+- Aaron Stone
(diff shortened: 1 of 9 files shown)


nshkrob:
Closing old PRs. (Also, this was fixed by eaed3e247b39d6d690).

sodabrew:
Awesome, thank you for implementing this upstream!

Issue #13, opened 2012-05-05
impossible to assign a default value to a bool


robey:
It looks like thrift doesn't define any syntax for true/false as a default value, so convention is to use 0 and 1, which may make sense to C people, but is objectively idiotic.

Scrooge should support both true/false and 0/1 as default values for bools.

Example from David Helder:

const bool ECHO_FRAMED = 1;

mosesn:
Looks like this was improved here: https://github.com/twitter/scrooge/commit/e9d2508703e22e5d4890af2ec216da71139afa93

PR #60, opened 2013-03-25
Warning for 'false' equals true

erikvanoosten:
Warn that 'false' is interpreted as true when it is used as a boolean literal. See also #13.


nshkrob:
Boolean constants handling was fixed in https://github.com/twitter/scrooge/commit/e9d2508703e22e5d4890af2ec216da71139afa93.

Issue #94, opened 2013-10-11
hashCode() for objects with optional fields throws IllegalArgumentException


eshelestovich:
Scrooge generates an incorrect `hashCode()` implementation for Java objects with optional fields causing `IllegalArgumentException`.

Steps to reproduce:
- Create a new Thrift structure:

```
struct TestStruct {
    1: required string key;
    2: optional binary value;
}
```
- Generate Java source.
- Create an instance of `TestStruct` class. Do not set value. 
- Call `hashCode()` directly or put the object into a `j.u.HashMap`:

``` java
final TestStruct obj = new TestStruct.Builder().key("key1").build();
new HashMap<Event, String>().put(event, "test");
```

Expected result: successfully calculated hash code.
Actual result:

```
java.lang.IllegalArgumentException
    at com.twitter.scrooge.Option$1.get(Option.java:23)
    at com.twitter.scrooge.Option$1.get(Option.java:21)
    at com.test.TestStruct.hashCode(TestStruct.java:100)
```

Fix #95

kevinoliver:
I believe this was for the experimental-java generated code which is no longer part of the project.

PR #95, opened 2013-10-11
Optional fields should not be included in hashCode() computation

eshelestovich:
This is a pull request for #94. 
We should probably exclude undefined optional fields from hashCode() computation.


mosesn:
This looks right to me.  Sorry it took so long for us to review it!  =_=

mosesn:
Actually, this patch is for the experimental-java scrooge, which no longer exists, so I'm going to close this.  Sorry about that!

PR #102, opened 2013-11-24
Fix typo

wtfleming:


mosesn:
LGTM, sorry this took so long for us to review it!

mosesn:
Actually, it looks like this was already fixed: https://github.com/twitter/scrooge/blob/develop/scrooge-generator/src/main/resources/scalagen/struct.scala#L229.  Sorry about that!

PR #103, opened 2013-12-09
Allow unions to contain fields with names of "primitive" types

willpugh:
Fix the ability for unions to contain fields with names of "primitive" types, e.g. Byte, Short, Int, Long, etc.

  *  The problem here was that when generating the union, the ScalaGenerator would create case classes with the names of the union's fields.  This would cause the declaration of a variable to use tthe local case classes instead of the scala.* versions. T
  *  This fix makes a trade off of a little more code in ths fix, for the benefit of having the generated code look better.  Mainly, instead of fully-qualifying all declarations, we only do it in a few spots.
  *  Repro case is added to naughty.thrift

### DIFF
--- a/project/build.properties
+++ b/project/build.properties
@@ -1,0 +1,1 @@
+sbt.version=0.12.1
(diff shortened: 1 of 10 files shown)


nshkrob:
Closing old PRs. Feel free to reopen with an updated patch.

Issue #99, opened 2013-11-15
ConstMap values for fields in a service function aren't supported.


seanparsons:
The IDL docs here:
http://thrift.apache.org/docs/idl/

Say that for a function in a service the syntax is the following:

```
'oneway'? FunctionType Identifier '(' Field* ')' Throws? ListSeparator?
```

For a field the syntax is as follows:

```
FieldID? FieldReq? FieldType Identifier ('= ConstValue)? XsdFieldOptions ListSeparator?
```

ConstValue allows the following:

```
IntConstant | DoubleConstant | Literal | Identifier | ConstList | ConstMap
```

A ConstMap is this:

```
'{' (ConstValue ':' ConstValue ListSeparator?)* '}'
```

Following on from that the following code should work fine (and does with the thrift command line tool): 

```
MyResponse myCall(1:FirstParam firstParam, 2:SecondParam secondParam = {"value": "cake"});
```

But the Scrooge generator gives the following error:

```
[error] (my-app-thrift/compile:scrooge-gen) com.twitter.scrooge.frontend.TypeMismatchException: Expecting StructType(Struct(SimpleID(SecondParam),SecondParam,List(Field(1,SimpleID(value),value,TString,None,Optional)),None),None), found MapRHS(List((StringLiteral(value),StringLiteral(cake))))
```

seanparsons:
Closing to cleanup my own list of issues.

PR #106, opened 2014-02-17
Generated Scala code now deals correctly with default values on structs and services.

seanparsons:
At the very least this should fix: http://github.com/twitter/scrooge/issues/99

In the process of doing so I've found a couple of other issues around defaults like false being set to true for booleans and the parser itself didn't handle boolean literals.


nshkrob:
Closing old PRs. Feel free to reopen with an updated patch.

PR #114, opened 2014-03-21
Upgraded scopt from 2.1.0 to 3.2.0

sslavic:
This patch contains
- scopt dependency upgrade from 2.1.0 to 3.2.0 and
- accompanying migration of scrooge-generator CLI argument parsing to new scopt

Old scopt version in scrooge, and new one in scalastyle-sbt-plugin v0.4.0 was causing issues like [this](https://github.com/scalastyle/scalastyle-sbt-plugin/issues/15)


nshkrob:
Scrooge is using scopt 3.2.0 now.

PR #135, opened 2014-08-13
Fix //-style comments on Windows

eirslett:
Don't match \n, but match \r?\n, so we can match Windows-style endlines \r\n.
Would be great if you could test this on Linux before merging (I only use Windows)


eirslett:
@kevinoliver could you please sync OSS? Would love to have this fix!

kevinoliver:
Not sure if you know, but its on the `develop` [branch](https://github.com/twitter/scrooge/commit/1a429b97bd52641338f4a151699fa423a3e812cc).

We'd need to release a new version to get it on to `master`, cc @travisbrown @nshkrob

PR #139, opened 2014-08-17
Generated Java code should read/write enums correctly

eirslett:
Sorry, I'm spamming you with pull requests now! :smile:

When reading/writing enums over the wire, they are defined
as TType.I32, and not TType.ENUM, as one would expect.
That quirk was already fixed in the generated Scala code.

- Port enum fix from Scala code generator to Java code generator


eirslett:
I'll close this PR and reopen it, so the PR is against the develop branch, not master.

Issue #129, opened 2014-06-09
Scala 2.11 support


arboreal84:
2.11 is the new stable version of the Scala language.
When will Scrooge support Scala 2.11?

adriancole:
Looks like we have an candidate fix https://github.com/twitter/scrooge/pull/198

Can someone see if this is ok to merge? This is blocking zipkin for example, who want to support scala 2.11

mosesn:
scrooge supports 2.11 successfully now!

PR #140, opened 2014-08-19
Scala 2.11 build for Scrooge-core

elbiczel:
As requested in #129 and discussed with @mosesn in twitter/finagle#290 here's the Scala 2.11 support for scrooge-core.

As in the ScalaTest migration guide: http://www.scalatest.org/user_guide/migrating_to_20 I've updated the ScalaTest to 1.9.2 for Scala 2.9.x and 2.10.x and as there were no deprecations updated the dependency to 2.2.1 (newest stable version) for Scala 2.11.x.

As soon as there is a finagle branch cross compiled for Scala 2.11 I'll be able to work on all the other parts of Scrooge to support Scala 2.11.

### DIFF
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -81,7 +81,7 @@ object Scrooge extends Build {
     otherResolvers += m2Repo,
 
     libraryDependencies ++= Seq(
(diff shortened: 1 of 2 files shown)


nshkrob:
Closing in favor of https://github.com/twitter/scrooge/pull/198/

PR #149, opened 2014-09-22
Make flag support type safe through the sbt plugin

jedesah:
This has the added advantage of bypassing the huge parsing step that didn’t ever really need to be called, so should make things altogether faster on top of avoiding weird behavior. Does it really make much sense to supply the “-help” in your sbt build?

### DIFF
--- a/scrooge-sbt-plugin/src/main/scala/com/twitter/ScroogeSBT.scala
+++ b/scrooge-sbt-plugin/src/main/scala/com/twitter/ScroogeSBT.scala
@@ -4,7 +4,7 @@ import sbt._
 import Keys._
 
 import java.io.File
-import scala.collection.JavaConverters._
+import com.twitter.scrooge.backend.{WithFinagle, ServiceOption}
 
 object ScroogeSBT extends Plugin {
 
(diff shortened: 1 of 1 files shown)


mosesn:
LGTM, let's see if we can merge this in!

mosesn:
This is extremely stupid, but adding some of these flags tips us over the edge to too many arguments for sbt.  It only [goes up to 11][0].  So I removed experimental flags because we don't seem to use them anywhere.

[0]: https://github.com/sbt/sbt/blob/1.0.x/main-settings/src/main/scala/sbt/Structure.scala#L374

jedesah:
@mosesn Do you want me to rebase this or are you going to do that as part of your internal merging process (twitter open source still works that way?)?

mosesn:
Yeah, I'll do it as part of the internal merging process, thanks!  It's up for review internally now.

mosesn:
This made it to `develop` here: https://github.com/twitter/scrooge/commit/5979724d4ac92f6b895847e0ebc7a5a67475fa85 thanks so much!  Really sorry about how long this took.

PR #150, opened 2014-09-24
Update version for 0.13.x

jedesah:
Follow up to:

http://stackoverflow.com/questions/19576513/what-is-the-repository-for-scrooge-sbt-plugin/26025979#26025979


jedesah:
I think this should be an easy merge at this point unless I am missing something.

nshkrob:
I think you need to change SBTPlugin.rst instead. The html file is generated from that by 'sbt scrooge-doc/make-site'.

nshkrob:
https://twitter.github.io/scrooge/SBTPlugin.html has been updated since, closing this.

PR #153, opened 2014-09-26
Fix #151

### LINKED ISSUE #151: [sbt-plugin] Add thrift files to published artifact
See this SO question for background: http://stackoverflow.com/questions/26026575/how-does-publishing-a-thrift-repo-work-with-scrooge-and-associated-plugins

jedesah:
[sbt-plugin] Add thrift files to published artifact

Problem

If someone has a project that compiles thrift but that also depends on a librairie that defines it's own thrift as well, this project needs to receive the thrift files of the dependency somehow. My understanding is that this functionality is part of the Scrooge maven plugin but not the sbt plugin currently.

Solution

The sbt-plugin now adds the thrift source files into the published jar file where both the maven plugin and the sbt plugin are already programmed to look for them.

Result

It is now possible for a project defining it's own thrift to depend on a "thrift" librairie using the Scrooge sbt-plugin to publish it's artifacts.

### DIFF
--- a/scrooge-sbt-plugin/src/main/scala/com/twitter/ScroogeSBT.scala
+++ b/scrooge-sbt-plugin/src/main/scala/com/twitter/ScroogeSBT.scala
@@ -208,8 +208,14 @@ object ScroogeSBT extends Plugin {
     sourceGenerators <+= scroogeGen
   )
 
+  val packageThrift = mappings in (Compile, packageBin) ++= {
+    (scroogeThriftSources in Compile).value.map { file =>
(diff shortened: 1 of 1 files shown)


mosesn:
We screwed up and got the wrong attribution data =(, but this was fixed by @jedesah here: https://github.com/twitter/scrooge/commit/1ab8b6650dddbf54f5d2bfa81a9321975ecfca3c

jedesah:
Ya, this one turned out to be a little controversial if I remember correctly. Not the end of the world if it wasn't attributed to me! 😛

PR #158, opened 2014-11-07
Scala TProcessor implementations for Some and Async

m50d:
Based on mirkonasato's branch, but I haven't received any reply from them.

Provides synchronous processors for services typed on `Some` (scalaz `Id` would be more correct but would introduce a dependency on scalaz at all times), and (when invoked with --scalaz) asynchronous processors for services typed on scalaz's `Task` (depending on https://github.com/apache/thrift/pull/253 for better async support in thrift ). These will depend on a little more at runtime; rather than make scrooge-core depend on scalaz I've created a scrooge-scalaz subproject that depends on scalaz, and the classes generated with --scalaz depend on this.

(Motivation: implementing a server without using Finagle - in particular because my server code needs to be scala 2.11. Also I like the scalaz `Task` abstraction.)

Arguably we should provide a processor that's generic in the same way as the interface is, but that's slightly beyond my current abilities. At the least, the distinction between sync and async seems pretty fundamental and hard to abstract over.

This is probably not an implementation that should be merged as-is - the code is not as abstracted as it should be, nor as well-tested, and support for oneway functions may be entirely missing. I'm submitting this as a pull request more to open a dialogue and check I'm moving in the right direction. Any advice about the design, or what kind of tests are necessary, would be very much appreciated.


nshkrob:
Hi @m50d, I'm going over the open scrooge PRs. Just briefly looking over this:
- Scrooge and Finagle are integrated fairly tightly, so I'm not sure the non-Finagle server fits into scrooge.
- scala 2.11 support should be coming soon in https://github.com/twitter/scrooge/pull/198/.

I'm closing this for now, feel free to reopen/comment if you are still interested in this.

m50d:
I am still interested - and professionally I'm still running my fork (in its https://github.com/VisualDNA/scrooge incarnation - I'm afraid I've mangled my work and personal accounts a bit here) in production[1]. From my perspective I'd like to write thrift servers in idiomatic, async scala style but without Finagle - it seemed like scrooge already supported non-Finagle clients? I'm not aware of any other projects that support an idiomatic scala style of thrift servers, and the vast majority of code to support the non-Finagle case was in common with scrooge. But obv. if you want to consider that to be out of scope then I'll look to other ways (either by continuing to maintain a fork or as a separate project)

[1] Awkwardly I'm about to leave VisualDNA but I don't yet know who will take over maintenance

nshkrob:
To update on this: we don't want to depend on scalaz because it is a big dependency.

Another thing: take a look at the ThriftMethod traits introduced in https://github.com/twitter/scrooge/commit/3fbb635936d82c00456b332647ea9136841f3227. It provides a generic description of a thrift method that we use to generate Finagle Services. If I understand the intent of this patch, some of this will be served by the ThriftMethods.

m50d:
FWIW I made a point of separating any scalaz dependency into its own module so that non-scalaz users would have no dependency on scalaz. If it's just scalaz that's the issue then I'd be willing (when I get time) to rework it to use stdlib futures (but if you don't want to support it either way then there's no value in doing that).

PR #167, opened 2015-01-12
[sbt-plugin] Fix overriding default settings

jedesah:
I'll be the first to admit that I am by no stretch of the mind an expert at sbt scopes, however I do think there is a problem and that this is a fix for it.

**Problem:**

After adding the Scrooge sbt-plugin with something like this:

`addSbtPlugin("com.twitter" %% "scrooge-sbt-plugin" % "3.17.1-SNAPSHOT")`

And having a build.sbt that looks something like this:

    ...

    com.twitter.scrooge.ScroogeSBT.newSettings

    scroogeBuildOptions := Seq()

    scroogeThriftSourceFolder <<= (baseDirectory) { _ / "thrift" / "wp-data" }

 None of these two settings overrides will be respected. The plugin will generate thrift with the defaults, that is Finagle bindings and look for thrift files in /scr/main/thrift

 **Solution:**

 Change scopes for default settings definitions so that one can easily change the defaults.

 **Result**

 Settings overrides are considered by the plugin.

### DIFF
--- a/scrooge-sbt-plugin/src/main/scala/com/twitter/ScroogeSBT.scala
+++ b/scrooge-sbt-plugin/src/main/scala/com/twitter/ScroogeSBT.scala
@@ -106,7 +106,7 @@ object ScroogeSBT extends Plugin {
    */
   val genThriftSettings: Seq[Setting[_]] = Seq(
     scroogeBuildOptions := Seq("--finagle"),
-    scroogeThriftSourceFolder <<= (sourceDirectory) { _ / "thrift" },
+    scroogeThriftSourceFolder <<= (sourceDirectory in Compile) { _ / "thrift" },
     scroogeThriftExternalSourceFolder <<= (target) { _ / "thrift_external" },
     scroogeThriftOutputFolder <<= (sourceManaged) { identity },
     scroogeThriftIncludeFolders <<= (scroogeThriftSourceFolder) { Seq(_) },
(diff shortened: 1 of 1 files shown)


jedesah:
Same here. "I think this change is looking pretty good now and we need this to get rid of our Scrooge fork. Can someone have one last look at this and merge it?"

mosesn:
LGTM, sorry it took so long for us to look at it!

mosesn:
Running into a problem while testing.  Looks like it can't parse this [file][0].  The reason seems to be because it can't find this relative include:

`Caused by: java.io.FileNotFoundException: finatra-thrift/finatra_thrift_exceptions.thrift`

I'm going to poke around, but I think the "in compile" part might not be the right thing for us.

[0]: https://github.com/twitter/finatra/blob/develop/examples/thrift-server/thrift-example-idl/src/main/thrift/calculator.thrift

jedesah:
Hmm... Ya, I really don't understand scopes very well. I remember that without this, "overriding" `SettingKey`s in the sbt build did not affect the result of printing out the current value of the setting key once the project was loaded. It's quite possible that there is some better way of doing this so that it works with all builds.

mosesn:
@jedesah from messing around with this, I think it works now.  I'm not sure what the state of the world was back when you submitted this patch, but it looks like you have to make your command

`scroogeThriftSourceFolder in Compile <<= (baseDirectory) { _ / "thrift" / "wp-data" }` and it'll be respected.

Again, I think this has changed since you last took a look, but it is now documented [here][0].

Do you think we can close this ticket?

[0]: http://twitter.github.io/scrooge/SBTPlugin.html#sbt-0-13-5-and-above

jedesah:
@mosesn If you say it works, I believe you!

mosesn:
@jedesah cool.  sorry again about how long this took.  we have really not been on top of scrooge for a while.

PR #169, opened 2015-02-05
Rewrite gen-test-thrift scripts to Scala

eirslett:
Problem:
The gen-test-thrift scripts (in scrooge-generator
and scrooge-benchmark) are OS-specific, so it's
not possible to build Scrooge on Windows. There is
also a lot of code duplication in the scripts.

Solution:
Rewrite the scripts into Scala and import them as
sbt tasks. This also gives us the power of Scala.


mosesn:
LGTM, @nshkrob what do you think?

project/ScroogeRunner.scala
```diff
@@ -0,0 +1,192 @@
...
+    Scala -> ""
+  )
+
+  type Input[T] = Types.Id[T]
+
+  class Runner(out: Input[Keys.TaskStreams], base: Input[File], cp: Input[Classpath], outputDir: Input[File]) {
+    def run(language: Language, namespace: String, finagle: Boolean = true, args: String) : Unit =  {
+      val finagleArg = if(finagle) "--finagle" else ""
+
+      val command = "java -cp " +
```
nshkrob: Scrooge is using scala 2.10 now, so feel free to use string interpolation in the future.

project/ScroogeRunner.scala
```diff
@@ -0,0 +1,192 @@
...
+      run(language = Scala, namespace = scalaVanillaNamespace, finagle = false, args = airportThriftFiles)
+    }
+
+    section("namespace/ with bar and java_bar as default namespace") {
+      val files = filesInDir("scrooge-generator/src/test/thrift/namespace") mkString " "
+      run(language = Java,
+        namespace = namespaces(Java) + "--default-java-namespace java_bar",
+        args = files)
+      run(language = Scala,
+        namespace = "" + "--default-java-namespace bar",
```
nshkrob: Remove the empty string

project/ScroogeRunner.scala
```diff
@@ -0,0 +1,192 @@
...
+      |-n thrift.def.default=thrift.apache_java_def.default_ns
+      |-n foo=apache_java_foo
+      |-n bar=apache_java_bar
+      |-n com.fake=com.apache_java_fake
+      |-n com.twitter.scrooge.integration_scala=com.twitter.scrooge.integration_apache_java
+      |--default-java-namespace apache_java_thrift
+    """.stripMargin
+
+  val namespaces = Map[Language, String](
+    Java -> javaNamespace,
```
nshkrob: You can put the namespace into the Language object.

nshkrob:
Cool, I think we'll get the detection of changes and continuous compilation of test files.

Please fix the two minor things above and we can merge this in.

eirslett:
Ok, I did some changes according to your feedback:
- move namespace into the Language objects
- remove empty string
- use string interpolation

nshkrob:
LGTM. Thanks, I'm pulling this in.

In the future, please add new commits after the review because it's easier to review changes incrementally. In this case there is just one commit, even though it includes changes after the review.

eirslett:
Ok, noted :-) thanks for merging it!

nshkrob:
Looks like this broke 'sbt scrooge-generator/test'. This is what I get on OS X at d4af76e38:

[error] /Users/nshkrob/code/scrooge-oss/scrooge-generator/src/test/scala/com/twitter/scrooge/integration/JavaIntegrationSpec.scala:6: object integration_apache is not a member of package com.twitter.scrooge
[error] import com.twitter.scrooge.{integration_apache => apacheGen}

I haven't gotten around to looking into this yet, but in case anyone has it'd be great to see the fix.

nshkrob:
Fixed in https://github.com/twitter/scrooge/pull/176

PR #173, opened 2015-02-25
Build things for scala 2.11

### LINKED ISSUE #172: My kingdom for a scrooge-generator_2.11 jar
All the dependencies seem there, but when I try to build with `./sbt ++2.11.5 package`, I get

```
[error] missing or invalid dependency detected while loading class file 'ScalaObjectHandler.class'.
[error] Could not access type ScalaObject in package scala,
[error] because it (or its dependencies) are missing. Check your build definition for
[error] missing or conflicting dependencies. (Re-run with `-Ylog-classpath` to see the problematic classpath.)
[error] A full rebuild may help if 'ScalaObjectHandler.class' was compiled against an incompatible version of scala.
```

And none exist on central, that I can see. I think my sbt-fu might just suck, help?

colinmarc:
I have no idea what I'm doing

mustache.java 0.9.0 requires java8, but that's the only version available for
scala 2.11. So I bumped everything to use java8.

Should fix #172.

### DIFF
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -17,6 +17,11 @@ object Scrooge extends Build {
   val utilVersion = "6.23.0" + suffix
   val finagleVersion = "6.24.0" + suffix
 
+  def isScala210x(scalaVersion: String) = scalaVersion match {
+      case version if version startsWith "2.10" => true
+      case _ => false
+  }
+
   def util(which: String) = "com.twitter" %% ("util-"+which) % utilVersion
   def finagle(which: String) = "com.twitter" %% ("finagle-"+which) % finagleVersion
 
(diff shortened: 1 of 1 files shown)


nshkrob:
Please pull the latest develop branch and try again - it'd be great to support 2.11.

Last I checked Scrooge on 2.11 needed mustache.java, which now exists for 2.11.
Our stack will likely be moving to JDK 8 soon, so that is not a blocker.

colinmarc:
Yeah, trying to rebase now, sorry. The problem now is that it's looking for the `SNAPSHOT` version of `util-codec`. Any ideas?

nshkrob:
You need to build and publish-local the dependencies, util and probably
finagle (see the e8ac94a27d1ab commit message).


    checkout util/develop, `./sbt publish-local`, then same for
finagle/develop.

colinmarc:
K, I'll try that. Also, the build on travis seems to be breaking independently of my changes.

colinmarc:
This works locally for me now, although I have to revert f1ef0cbbfcff7c7041988ebb84f03704d842330c, because it doesn't seem to be building my thrift on OSX.

nshkrob:
Closing in favor of https://github.com/twitter/scrooge/pull/198.

PR #176, opened 2015-03-03
Fix scrooge-generator/test

nshkrob:
Problem

sbt scrooge-generator/test was broken because of the namespace mismatch.

Solution

Fix the namespace.

Note: change to Build.scala is to build on travis. We had some changes to how we build stuff locally which doesn't work on travis ATM.


nshkrob:
This was merged.

PR #178, opened 2015-03-18
Fix scrooge gen time file triggers

nitay:
There's a bug in scrooge sbt plugin - if it is used with anything else that generates scala files it doesn't get the right timestamp. The problem is it looks at timestamp from *all* scala files generated. So for example when using with Play (which we do), often Play will run first, generate some route file, and then scrooge thinks it has run b/c the timestamp of the Play scala file is newer than the thrift file.

The fix is to generate a specific file when scrooge runs, and look at its timestamp.

Works for me, LMK what you guys think.

### DIFF
--- a/scrooge-sbt-plugin/src/main/scala/com/twitter/ScroogeSBT.scala
+++ b/scrooge-sbt-plugin/src/main/scala/com/twitter/ScroogeSBT.scala
@@ -4,6 +4,7 @@ import sbt._
 import Keys._
 
 import java.io.File
+import java.util.Date
 import scala.collection.JavaConverters._
 
 object ScroogeSBT extends Plugin {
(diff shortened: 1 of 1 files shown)


nshkrob:
Its true that the timestamp detection would fail in the case you described, but using the special timestamp file looks brittle to me.

To avoid collisions with other tools, you can tell the plugin to put generated files to a different directory, e.g.:

`com.twitter.scrooge.ScroogeSBT.scroogeThriftOutputFolder in Compile := sourceManaged.value / "scrooge-generated-files"`

nitay:
I tried that approach but we found it a pain to use with IDEs and other tools that expect the standard layout.

nshkrob:
Note that with the suggested approach the generated files still end up in a subdirectory of target/scala-2.10/src_managed/, so I'd expect IDEs to find them. Can you try this?

PR #179, opened 2015-03-19
Documentation update on how to get scrooge-sbt-plugin.

kristofa:
The sbt plugin was not added to the aggregates. This was confusing
as after publish it was not available.

### DIFF
--- a/doc/src/sphinx/SBTPlugin.rst
+++ b/doc/src/sphinx/SBTPlugin.rst
@@ -1,6 +1,8 @@
 SBT Plugin
 ==========
 
+The scrooge-sbt-plugin is not published to Maven Central but published to `Bintray <https://bintray.com/twittercsl/sbt-plugins/scrooge-sbt-plugin/view>`_
+
 Add a line like this to your `project/plugins.sbt` file:
 
 ::
--- a/doc/src/sphinx/index.rst
+++ b/doc/src/sphinx/index.rst
@@ -63,6 +63,8 @@ To build scrooge, use sbt:
 
     $ ./sbt +publish-local
 
+This will currently not build and publish the scrooge-sbt-plugin.
+
 User's guide
 ------------
 
--- a/doc/src/sphinx/index.rst
+++ b/doc/src/sphinx/index.rst
@@ -61,7 +61,7 @@ To build scrooge, use sbt:
 
 ::
 
-    $ ./sbt +publish-local
+    $ ./sbt +publishLocal
 
 This will currently not build and publish the scrooge-sbt-plugin.
 You can still build the scrooge-sbt-plugin separately by executing:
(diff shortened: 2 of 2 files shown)


mosesn:
My impression was that we publish the sbt plugin separately from everything else.  Maybe it would be better to update the docs?

nshkrob:
LGTM.

@mosesn You are probably thinking of the maven plugin. That's separate because it's built with maven.

mosesn:
@nshkrob I thought the sbt plugin was published to bintray, no?

nshkrob:
We've talked with @mosesn and it looks like doing this would publish the scrooge-sbt-plugin to Maven Central as a maven artifact. That's bad because sbt plugins should include scala and sbt versions, so they must be published as ivy artifacts, not maven. We are publishing the scrooge-sbt-plugin to bintray.com in ivy format. 

One way around this is to write a task to remove scrooge-sbt-plugin from publishing to Maven Central.

kristofa:
Thanks for the quick response. I see that the sbt plugin can't be published to Maven Central as in that case all dependencies should be in Maven central which is not the case.

It would take me some time to find out how to change the sbt configuration to let public-local work but avoid deploying/publishing to Maven Central.  It might be easier to update the documentation to indicate scrooge-sbt-plugin is published and available at bintray and won't be published locally by default?

mosesn:
I'd be happy with that solution.

kristofa:
I added two commits to the documentation instead. Please have a look. Thanks!

mosesn:
Could you mention that you can still publish the sbt plugin by manually going to that project and using publish-local?  it should be something like:

```
project scrooge-sbt-plugin
publish-local
```

travisbrown:
This is a tiny point, but can we use the camel case `publishLocal` in any new documentation?

SBT is phasing out kebab: `publish-local` still works for now but doesn't show up in autocomplete, etc.

kristofa:
@mosesn @travisbrown I did some more commits based on your remarks. I also verified that the instructions work.

travisbrown:
@kristofa Thanks! One other small point: SBT 0.13 itself always uses Scala 2.10 (whether it's being used to build 2.9, 2.10, or 2.11 projects), so SBT plugins are always only published for a single Scala version, and the `+` on `+publishLocal` doesn't do anything.

It's not a big deal, but it's potentially a little confusing—do you mind removing it?

kristofa:
@travisbrown I removed the cross building and used regular `publishLocal`.

mosesn:
LGTM

mosesn:
This made it to develop here: https://github.com/twitter/scrooge/commit/d0d9cbbbbb5c351d9b570cddf136568c8906128f !  Thanks =)

PR #180, opened 2015-03-24
Upgrade sbt plugin to be an auto plugin

benjumanji:
- Allow for usage simply by including the plugin in the project.
- Inject settings automatically.


mosesn:
This is cool, but it seems like a pretty breaking change.  In particular, I'm worried about newSettings, which will disappear without giving folks a useful error message.  Maybe we should deprecate newSettings instead?  Also, can you update the docs to explain how to use the autoplugin?

benjumanji:
Excellent points! I will fix that.

travisbrown:
Agreed, @benjumanji, this is very cool. Would you mind also rebasing against `develop`? There's a new key and [some other small changes](https://github.com/twitter/scrooge/commit/1ab8b6650dddbf54f5d2bfa81a9321975ecfca3c) there since `master`.

benjumanji:
Question, when we say update the docs, are we talking in the gh-pages site or the README? Or both?

benjumanji:
`newSettings` now has a deprecation warning, and rebased against develop. Should I close this and point at develop for the PR?

I'll add some docs as soon as I know where to put them.

benjumanji:
Something like the migration [guide](https://github.com/sbt/sbt-assembly/blob/master/Migration.md) for assembly will probably be helpful, as that underwent a similar transition recently.

benjumanji:
`com.twitter#finagle-thrift_2.10;6.24.0-SNAPSHOT` doesn't seem to be published on central.

travisbrown:
@benjumanji Starting a new PR against `develop` sounds reasonable to me. Also you could define `newSettings` in terms of `projectSettings` to avoid a deprecation warning (e.g. like [this](https://github.com/travisbrown/scrooge/commit/ef073ad525eae8be13ce849ff32c6848d4013d1e)).

I just tried this out on a small example project after publishing Finagle locally from `develop` and it seems to work just fine (after the `scroogePublishThrift in Compile` change in my commit).

@nshkrob This isn't related to this PR specifically, but was this actually working before after [this change](https://github.com/twitter/scrooge/commit/1ab8b6650dddbf54f5d2bfa81a9321975ecfca3c#diff-3d1ecf501419eb7452297ced4c03b47eR213)? I'm getting errors that `*:scroogePublishThrift` isn't defined, which makes sense—it seems like it also needs the `in Compile` modifier, right?

PR #181, opened 2015-03-24
Upgrade sbt plugin to be an auto plugin

benjumanji:
- Allow for usage simply by including the plugin in the project.
- Inject settings automatically.

Continuation #180 for those following.

- [x] Actually test the syntax in the Migration guide

### DIFF
--- a/scrooge-sbt-plugin/src/main/scala/com/twitter/ScroogeSBT.scala
+++ b/scrooge-sbt-plugin/src/main/scala/com/twitter/ScroogeSBT.scala
@@ -5,7 +5,7 @@ import Keys._
 
 import java.io.File
 
-object ScroogeSBT extends Plugin {
+object ScroogeSBT extends AutoPlugin {
 
   def compile(
     log: Logger,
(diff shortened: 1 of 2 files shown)


travisbrown:
Sorry to keep dragging this out, @benjumanji, but I think it'd be best to put the migration docs (which look awesome) [here](https://twitter.github.io/scrooge/SBTPlugin.html) in order to avoid spreading things out all over the place. (At least this is the approach we decided on for Finagle—"meta" stuff like the adopters list and contributor guidelines go in the GitHub Markdown docs, while everything about usage goes in the Sphinx site.)

I'd be happy to move your content over and change the formatting from MD to RST. I really appreciate you doing this.

benjumanji:
@travisbrown no worries! I am very happy to do it. Would you like any other docs? It would be nice if I could get a build of this running locally so I can test out the suggested migrations before anyone actually tries them. Is that finagle thrift snapshot published anywhere?

travisbrown:
@benjumanji No, unfortunately we're not currently publishing Finagle snapshots (I'm hoping we can start soon), so you'll have to grab the repo and `publishLocal` from the `develop` branch.

I don't think we need any other docs, but @nshkrob should have the final word on that.

benjumanji:
@travisbrown Cool. I have re-written the docs, added them to the gh-pages branch so expect a separate PR on that. I'll drop the .md from this PR and will push it all in a few hours once I have tripled-checked my spelling and got things tested locally.

Thanks for the feedback.

travisbrown:
Oh, sorry I wasn't clear! The `gh-pages` site is created from the Sphinx site in [the `doc/src` directory](https://github.com/twitter/scrooge/tree/develop/doc/src/sphinx). If you update [`SBTPlugin.rst`](https://github.com/twitter/scrooge/blob/develop/doc/src/sphinx/SBTPlugin.rst) there your changes will get pushed to `gh-pages` the next time we publish a release.

benjumanji:
ahh, ok. I'll fix that.

benjumanji:
Ok, docs amended, commits squashed, syntax tested, ready for your consideration. Feel free to nit anything in the docs, I am happy to adjust.

scrooge-sbt-plugin/src/main/scala/com/twitter/ScroogeSBT.scala
```diff
@@ -210,13 +221,16 @@ object ScroogeSBT extends Plugin {
   )
 
   val packageThrift = mappings in (Compile, packageBin) ++= {
-    (if (scroogePublishThrift.value) (scroogeThriftSources in Compile).value else Nil).map { file =>
+    (if ((scroogePublishThrift in Compile).value) (scroogeThriftSources in Compile).value else Nil).map { file =>
```
mosesn: I'm not super familiar with the sbt dag stuff, does forcing the value of scroogePublishThrift make it harder for sbt to compile the dag correctly?  Isn't there a fishbone operator that we can use for that kind of thing?

travisbrown: The `.value` macro will be expanded into something involving `<<=` (or related operators—I'm not sure about the details). In any case looking up this value in a `mapping in …` task shouldn't make anything harder for SBT.

mosesn: Yeah, <<= was what I was thinking.  If value is a macro, we should be set.

mosesn:
LGTM

travisbrown:
@benjumanji Do you mind pulling in [this commit](https://github.com/travisbrown/scrooge/commit/0477de88ed7694377e2cf25ff9ea45d8c44271dc) from my fork with some small changes to wording, etc., if it looks okay to you?

After that (and @nshkrob's sign-off on the `(scroogePublishThrift in Compile).value` change) this is :+1: from me.

benjumanji:
@travisbrown :+1: for wording changes. I squashed the commit in directly, if you'd like the changes split out, lemme know.

travisbrown:
@benjumanji All the changes will get squashed to a single commit when this gets merged internally, so :+1:, thanks.

nshkrob:
I'm catching up on this review - the scroogePublishThrift change looks good. @travisbrown, you are right, it wasn't working without it. It would be good to have a test for scrooge-sbt-plugin (publish it and build a test project with it).

nshkrob:
LGTM, thanks! I'll pull this in.

mosesn:
u da :bomb:

travisbrown:
This is now [on GitHub](https://github.com/twitter/scrooge/commit/753e787af57664b9f1c795211017a051d82d8982) and available in 3.18.0. Thanks again, @benjumanji!

PR #183, opened 2015-04-27
Update docs with new versions and repository for SBT plugin

travisbrown:
Problem

Because we've reverted libthrift to 0.5.0-1, it's now necessary to add the
Twitter Maven repository to the SBT meta-meta-build for projects that use the
Scrooge SBT plugin. The docs don't make this clear.

Solution

Add a line to the example in the docs. I've also updated some other versions
while I'm here.

### DIFF
--- a/doc/src/sphinx/SBTPlugin.rst
+++ b/doc/src/sphinx/SBTPlugin.rst
@@ -5,6 +5,8 @@ Add a line like this to your `project/plugins.sbt` file:
 
 ::
 
+    resolvers += "Twitter's Repository" at "http://maven.twttr.com/"
+
     addSbtPlugin("com.twitter" %% "scrooge-sbt-plugin" % "3.18.0")
 
 In your `build.sbt` file:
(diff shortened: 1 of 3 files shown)


williammmeyer:
+1

This fixes a related issue with the docs:
scrooge-generator 3.18 creates code that depends on twitter.util 6.24
finagle-core 6.24 depends on twitter.util 6.23

    [error] .../src_managed/.../thrift/TaskService$FinagleClient.scala:13: object Throwables is not a member of package com.twitter.util
    [error] import com.twitter.util.{Future, Return, Throw, Throwables}

https://github.com/twitter/scrooge/blob/scrooge-3.18.0/project/Build.scala#L17
https://github.com/twitter/scrooge/blob/scrooge-3.18.0/scrooge-generator/src/main/resources/scalagen/finagleClient.scala#L7
https://github.com/twitter/finagle/blob/finagle-6.24.0/project/Build.scala#L11

nshkrob:
LGTM. @travisbrown do you want to submit this internally?

doc/src/sphinx/SBTPlugin.rst
```diff
@@ -5,6 +5,8 @@ Add a line like this to your `project/plugins.sbt` file:
 
 ::
 
+    resolvers += "Twitter's Repository" at "http://maven.twttr.com/"
```
taylorleese: I'm sure there are various instances of this around, but it should move to https now.

mosesn:
LGTM

mosesn:
Sorry, I misunderstood what this patch was doing.  I think we've already updated all of these, so we should be set.  Sorry about that!

PR #113, opened 2014-03-17
add language option to scrooge-sbt-plugin

thirstycrow:

PR #187, opened 2015-05-04
scrooge-sbt-plugin: Add generated source option

benjumanji:
Problem:

scala is hardcoded into the plugin, even though sbt will
happily compile java code. This is an unnecessary limitation.

Solution:

Add a setting key for selecting the language, and calculate
the generated source extension from that.

### DIFF
--- a/scrooge-sbt-plugin/src/main/scala/com/twitter/ScroogeSBT.scala
+++ b/scrooge-sbt-plugin/src/main/scala/com/twitter/ScroogeSBT.scala
@@ -36,9 +36,36 @@ object ScroogeSBT extends AutoPlugin {
 
   object autoImport {
 
+    sealed trait ScroogeGeneratedSource
+
+    object ScroogeScala extends ScroogeGeneratedSource {
+      override def toString(): String = "scala"
+    }
+
+    object ScroogeJava extends ScroogeGeneratedSource {
+      override def toString(): String = "java"
+    }
+
+    object ScroogeExperimentalJava extends ScroogeGeneratedSource {
+      override def toString(): String = "experimental-java"
+    }
+
+    private[scrooge] def globOf(language: ScroogeGeneratedSource): String = {
+      language match {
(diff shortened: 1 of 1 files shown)


benjumanji:
I am not sure if this would be better if we pulled this out the build options setting, but given that langauge is already being treated specially in the call to the compiler, this doesn't feel weird.

```
scroogeGeneratedSource in Compile := ScroogeExperimentalJava
```

is the intended usage.

vkostyukov:
Hi @benjumanji!

First of all, thanks a lot for this PR!

We've faced this issue recently and started working on the fix. In fact, there is already an internal PR that does the same thing but in different manner. It's also taking into account the issue with class-loaders and mustache templates (see https://github.com/twitter/scrooge/pull/113 for details), which, I think, makes it a correct solution.

That's said, our PR should go alive today. And, I think, we should probably merge our patch in favour of this one since it's already tested and reviewed internally.

Sorry for this. And thanks again!

benjumanji:
No worries, all I care about is getting some sweet java code generated! Do you have any idea when this will make it into a public release?

benjumanji:
ping @vkostyukov 

Would appreciate some indication of timeline, mostly around whether I can afford to wait until this is published or if it is worth the hassle of republishing a fork internally to use until the change is public.

vkostyukov:
Sorry @benjumanji for the delay.  Our change has landed in master yesterday. Unfortunately, I don't have any estimations on when it will show up here or even be released. Maybe @nshkrob can give you some details.

nshkrob:
The plugin change is now pushed to develop. 

As for a versioned release, you should probably expect one within a month from now (we don't have a set date though).

vkostyukov:
@nshkrob do we publish shapshots by any chance?

nshkrob:
No, never got around to setting it up, although it would be useful.

benjumanji:
Thanks! I will republish develop (better than maintaining my own changes).

PR #189, opened 2015-05-06
Changes `scroogeGen` to only return the files it generated.

### LINKED ISSUE #188: `scroogeGen` list of generated source files is wrong
The list of generated source files that `scroogeGen` in the Scrooge SBT plugin returns contains all the files under `src_managed` instead of just the files added by Scrooge. This means that any existing files are added twice.

Using Scala 2.11 this causes the compiler to fail with the error `x is already defined as ...`.

stephanh:
Previously `scroogeGen` would return all files under `src_managed` this
breaks the Scala 2.11 compiler if another generated also generated files
and returned them.

This uses a mutable ListBuffer in a setting to cache the list of files
that have been generated.

This closes #188.

### DIFF
--- a/scrooge-sbt-plugin/src/main/scala/com/twitter/ScroogeSBT.scala
+++ b/scrooge-sbt-plugin/src/main/scala/com/twitter/ScroogeSBT.scala
@@ -1,5 +1,7 @@
 package com.twitter.scrooge
 
+import scala.collection.mutable.ListBuffer
+
 import sbt._
 import Keys._
 
@@ -15,7 +17,7 @@ object ScroogeSBT extends AutoPlugin {
     namespaceMappings: Map[String, String],
     language: String,
     flags: Set[String]
-  ) {
+  ): List[String] = {
     val compiler = new Compiler()
     compiler.destFolder = outputDir.getPath
     thriftIncludes.map { compiler.includePaths += _.getPath }
@@ -100,6 +102,12 @@ object ScroogeSBT extends AutoPlugin {
       "scrooge-gen",
       "generate scala code from thrift files using scrooge"
     )
+
+    // Keeps track of the generated scala files so we can cache this list without having to return scroogeGen
+    val scroogeGenerated = SettingKey[ListBuffer[File]](
+      "scrooge-generated",
+      "generated scala files by scrooge"
+    )
   }
 
   import autoImport._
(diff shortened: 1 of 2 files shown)


stephanh:
Has anyone had a chance to look at this?

nshkrob:
Thanks for the patch;

What happens if the files are already generated and don't need to be rebuilt? (The logic of scroogeIsDirty)

stephanh:
It hasn't really changed. Where before it would check all scala files in the output folder it now only checks against the Scala files it generated.

scrooge-sbt-plugin/src/main/scala/com/twitter/ScroogeSBT.scala
```diff
@@ -179,8 +188,9 @@ object ScroogeSBT extends AutoPlugin {
       streams,
       scroogeThriftSources,
       scroogeThriftOutputFolder,
-      scroogeThriftIncludes
-    ) map { (out, sources, outputDir, inc) =>
+      scroogeThriftIncludes,
+      scroogeGenerated
```
nshkrob: Looks like this is initialized to the empty list so scrooge will always regenerate thrift, i.e. this patch breaks the dirty check.

The problem is that it doesn't know what the output files will be.

One workaround is to generate output in a separate directory, so no other files are there.

stephanh: It will be updated as soon as some files have been generated by scroogeGen. Are you concerned about files being recreated across different runs of sbt?

nshkrob: Yes, that's what I mean. Now it will always regenerate files on the first run, which is incorrect.

stephanh: How do you ensure that if you change the sbt settings it regenerates the files using the old approach?

Apart from having the behaviour it already had and writing to a different directory or writing out some state I'm not sure how we can achieve what you want?

nshkrob: The current check is simple - check all the *.scala files in the output directory (or .java - support was added in cd5e6c0b6f3e0).

You could figure out the output files by parsing the input and figuring out what structs and services would be generated. This is more correct but too complex. I don't know if there is a good solution for this.

nshkrob:
Closing old PRs (feel free to reopen with an updated patch).

PR #193, opened 2015-06-15
Fixing mustache dependency for Scala 2.11 cross version support

csaltos:
Mustache support for Scala 2.11 is included in the 0.9.0 version but it's Java 8 only so use the 0.8.17 that is Java 7 in the meantime.

Currently the mustache 0.8.17 is only available as scala-extensions-2.10 but it seems to be compatible with Scala 2.11 and especially with Java 7.

A request for mustache 0.8.x for scala-extensions-2.11 is asked at https://groups.google.com/forum/#!topic/mustachejava/8acnzuaoAG8

### DIFF
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -58,8 +58,7 @@ object Scrooge extends Build {
   val sharedSettings = Seq(
     version := libVersion,
     organization := "com.twitter",
-    // 2.11-ification needs more work with mustache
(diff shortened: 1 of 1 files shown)


csaltos:
Travis is missing finagle and util dependencies, so creating a new PR trying to correct this.

csaltos:
Fixing old dependencies in this PR -> https://github.com/twitter/scrooge/pull/194

PR #194, opened 2015-06-15
Enabling Scala 2.11 crossversion for Scrooge

csaltos:


mosesn:
Given that scala 2.11 is not backwards compatible with 2.10, I really don't want to depend on a 2.10 lib in scrooge.  For one thing, even if it doesn't rely on *any* features that have been removed, it'll slurp in the scala 2.10 lib, as well as the 2.11 lib, neither of those is exactly slim.

We're probably going to be moving to jdk8-only, so I think it's wiser to just wait until that happens and upgrade the mustache dependency.  If you're having a problem in the mean time, it might make sense to temporarily use a fork (if you really have to) or use something like jarjar to avoid the dependency collision.

johnynek:
Also, this

johnynek:
Sorry, fat finger...

csaltos:
Yes, it makes sense. Let's wait until we move to JDK 8. What's the rush.

By the way I think the SNAPSHOP fixes for util and finagle are still required for a successful build, especially at TravisCI. I will try to create a separate PR with that fix.

Thank you for your guidance.

Regards,

Carlos Saltos

csaltos:
Ready !! ... I put the build fix in a separate PR here -> https://github.com/twitter/scrooge/pull/195 I hope it's useful.

PR #195, opened 2015-06-16
Fix TravisCI builds

csaltos:
SNAPSHOT versions not available for finagle nor util so using finagle-6.25.0 and util-6.24.0 directly.


mosesn:
@csaltos the reason why we have these is so that scrooge `develop` users can still compile when they're using features from `util` or `finagle` that haven't been published to maven central yet.  Maybe the right solution is to publish-local `util` and `finagle` instead?

mosesn:
It's also a good point that it's hard to tell why the `+ suffix` bit has been added.  If you add a comment explaining about depending on `develop` versions of `util` and `finagle`, we could probably merge that in too.  Thanks for your contributions, we really appreciate it :+1:

csaltos:
OK, the suffix trick is actually a good trick, but since neither util-6.24.0-SNAPSHOT nor finagle-6.25.0-SNAPSHOT are available Travis CI is constantly reporting errors for missing those dependencies.

I will add the comment you mention, I hope I'm able to explain it well.

csaltos:
OK, ready, I hope it's correct.

mosesn:
Do you want to try to fix the build here too: https://github.com/twitter/scrooge/blob/master/.travis.yml#L10 or just do the comment bit in this PR?

csaltos:
I will try to fix the build at https://github.com/twitter/scrooge/blob/master/.travis.yml#L10 too

project/Build.scala
```diff
@@ -14,8 +14,12 @@ object Scrooge extends Build {
   val suffix = if (branch == "master") "" else "-SNAPSHOT"
 
   val libVersion = "3.18.1" + suffix
-  val utilVersion = "6.24.0" + suffix
-  val finagleVersion = "6.25.0" + suffix
+
+  // For develop users: if you are using new features in util and finagle dependencies
+  // that are not yet published to Maven Central, you can publish util and finagle locally
+  // or you can add the "+ suffix" to get the SNAPSHOT versions accordingly
+  val utilVersion = "6.24.0"
```
mosesn: actually, could you leave it so that it still says "+ SNAPSHOT"?

mosesn: *+ suffix

csaltos: OK

csaltos: done

csaltos:
Now scroog build is broken at Travis CI but I will try to fix it from .travis.yml

csaltos:
OK, I think I fix the build at .travis.yml but I mix the PR with other commits, I'm sorry, I will create another PR with a clean commit

csaltos:
OK, I move the Travis CI fix to the PR https://github.com/twitter/scrooge/pull/196

PR #196, opened 2015-06-16
scrooge: Fixing Travis CI builds

csaltos:
Problem

Travis CI builds are broken because they are looking for SNAPSHOT versions of their dependencies and they are not found.

Solution

Deploy the SNAPSHOT versions locally cloning the remote repos and then using sbt publishLocal

Result

Now Travis CI builds are working using the locally published SNAPSHOT versions of their dependencies


csaltos:
Ready @mosesn the build is now fixed at .travis.yml for non master branches by publishing local SNAPSHOT versions accordingly.

build-support/bin/ci.sh
```diff
@@ -0,0 +1,61 @@
+#!/bin/bash
+
+function clone_repo {
+  local name=$1
+  local repo=$2
+  local branch=$3
+  git clone $repo || {
```
mosesn: we can simplify this by using `git clone "$repo" --branch "$branch"`

build-support/bin/ci.sh
```diff
@@ -0,0 +1,61 @@
+#!/bin/bash
+
```
mosesn: set -e?

build-support/bin/ci.sh
```diff
@@ -0,0 +1,61 @@
...
+    echo "Unable to publish local $name with target $target"
+    exit 5
+  }
+  cd ..
+}
+
+function bootstrap_scrooge {
+  local current_branch=$(git rev-parse --abbrev-ref HEAD)
+  if [ "$current_branch" != "master" ]; then
+    mkdir -p build-support/tmp
```
mosesn: maybe use mktemp instead?

build-support/bin/ci.sh
```diff
@@ -0,0 +1,61 @@
+#!/bin/bash
```
mosesn: I don't love putting this in build-support/bin, we already have a bin directory, can we use that?

build-support/bin/ci.sh
```diff
@@ -0,0 +1,61 @@
...
+}
+
+function publish_local {
+  local name=$1
+  local target=$2
+  cd $name || {
+    echo "Unable to get $name for publish local $target"
+    exit 4
+  }
+  ../../../sbt ++$TRAVIS_SCALA_VERSION $target || {
```
mosesn: why not use ./sbt?

build-support/bin/ci.sh
```diff
@@ -0,0 +1,61 @@
...
+    echo "Unable to checkout branch $branch for $name from $repo"
+    exit 3
+  }
+  cd ..
+}
+
+function publish_local {
+  local name=$1
+  local target=$2
+  cd $name || {
```
mosesn: This seems unlikely to fail, and if it does, `set -e` will help us out with it.  I don't think we need the || check.

build-support/bin/ci.sh
```diff
@@ -0,0 +1,61 @@
...
+    echo "Unable to get $name for publish local $target"
+    exit 4
+  }
+  ../../../sbt ++$TRAVIS_SCALA_VERSION $target || {
+    echo "Unable to publish local $name with target $target"
```
mosesn: this is cool, but we could probably just publish-local everything.  do you not want to do that because it's slow?

mosesn:
This is very cool, but I think if we use set -e we can avoid a lot of the manual error checking, since the tools themselves can explain what went wrong.  We could potentially add -x too.

What do you think?

csaltos:
Thank you for tour review !! :+1: :smile: ... I'll reply on every comment of yours

csaltos:
Yes, using set -e and even with -x will be a lot better, I will do that change.

build-support/bin/ci.sh
csaltos: Yes, it was because it was slow, but on top of that it's because a couple of finagle projects got errors while building from develop, so I try to reduce the impact on finagle dependencies including the ones that specifically scrooge depends on. What do you think ?

build-support/bin/ci.sh
csaltos: Yes, good point, will change it with set -e and -x too.

build-support/bin/ci.sh
csaltos: using the ./sbt included with finagle and util gives errors, exactly the error is ->

bad sbt-launch.jar.  delete sbt-launch.jar and run ./sbt again.

I tried to delete de sbt-launch.jar just for curiosity but the error remains.

For a moment I though it was a local error on my machine but Travis CI gives the same error.

So, I just use the sbt included on scrooge.

Do you have the same error using ./sbt from finagle ?

I will try to figure out what the error is and we can discuss about it.

build-support/bin/ci.sh
csaltos: Yes, good point. I will put it in the already existing bin directory.

build-support/bin/ci.sh
csaltos: Duh for me !! ... you are right, that's even better (easier and cleaner). I will change it to mktemp instead.

build-support/bin/ci.sh
csaltos: On the way !!

build-support/bin/ci.sh
csaltos: That's right ... I will go for it and change it at the script.

build-support/bin/ci.sh
mosesn: Hmm, that's very strange.  That should only happen if you're in a directory which already has an sbt-launch.jar.  Don't spend too much time on it, if you can't figure out why, just add a TODO.

csaltos: I've found the error -> The SBTOPTS from Travis CI are clashing with the internal ./sbt

So I just added unset SBTOPTS to the .travis.yml file and it works now and actually finagle's .travis.yml file has the save unset SBTOPTS configuration.

csaltos: Oh oh !! ... the ./sbt of util is crashing at the JVM when using openjdk7 and scrooge sbt is working just fine. I will put a TODO by now.

csaltos:
Ready !! :+1: :smile: ... I've added the changes you proposed and the script is now cleaner and to the point.
I've just leaved a TODO for the sbt part because the internal ./sbt of the other projects are breaking the JVM when using openjdk7 with a buffer overflow error. I will try to solved now.

mosesn:
LGTM!  Can you add a commit in the format described here: https://github.com/twitter/finagle/blob/master/CONTRIBUTING.md#style so it's easy to reason about when we merge it in?  We should really have a CONTRIBUTING.md for scrooge too, but looks like we forgot to include one.

@nshkrob @travisbrown could you guys sign off on this too?

csaltos:
Ready Moses, now I added a commit in the format you refer me. If there is something else to complete this PR, please let me know.

project/Build.scala
```diff
@@ -14,6 +14,10 @@ object Scrooge extends Build {
   val suffix = if (branch == "master") "" else "-SNAPSHOT"
 
   val libVersion = "3.18.1" + suffix
+
+  // For develop users: if you are using new features in util and finagle dependencies
```
travisbrown: Can we say "For users of the develop branch"? Seems a little clearer.

travisbrown: Also, it might be a good idea to recommend publishing Util and Finagle locally for _anyone_ working on the develop branch—users may not have any idea whether Scrooge is using features that have been introduced since the last release, and if they guess wrong they may end up with weird errors (possibly at runtime).

csaltos: Of course. Here it goes !!

nshkrob:
LGTM, thanks for the patch!

travisbrown:
This looks great! Thanks much, @csaltos!

csaltos:
You are very welcome Moses and Travis !! :+1: :smile:

project/Build.scala
csaltos: Ostrich is also a required local dependency when working on the develop branch.

csaltos: Anyway, I think that explanation is already included at https://github.com/twitter/scrooge/blob/master/README.md

csaltos:
The previous build failed and is because sometimes the builds got network errors while downloading external artifacts.

If it's OK with you, I will try to activate cache for this external artifacts to keep the builds better and also to make them faster. It's something similar to the cache configuration at Finatra here: https://github.com/twitter/finatra/blob/master/.travis.yml

csaltos:
Ready, the cache for external artifacts is now ready.

nshkrob:
I've pulled this in, should show up on develop in a few days. Thanks for the patch!

csaltos:
Great, thank you Nik !!

PR #198, opened 2015-07-18
Upgrade Scrooge to support Scala 2.11

eirslett:
Problem:
Scrooge only compiles with Scala 2.10, but many
projects need Scrooge on 2.11.

Solution:
Make small changes to the code, so that it works
with Scala 2.11.

- Use stable versions of Finagle (since this is the OSS version)
- Move scrooge version to version.sbt
- Inline Mustache.java's Scala extension
- Bump junit

### DIFF
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -64,7 +64,7 @@ object Scrooge extends Build {
     version := libVersion,
     organization := "com.twitter",
     // 2.11-ification needs more work with mustache
-    crossScalaVersions := Seq("2.10.5" /*, "2.11.6"*/),
+    crossScalaVersions := Seq("2.10.5" , "2.11.7"),
     scalaVersion := "2.10.5",
 
     resolvers ++= Seq(
@@ -93,7 +93,7 @@ object Scrooge extends Build {
     libraryDependencies ++= Seq(
       "org.scalatest" %% "scalatest" % "2.2.4" % "test",
       "org.scalacheck" %% "scalacheck" % "1.12.2" % "test",
-      "junit" % "junit" % "4.10" % "test" exclude("org.mockito", "mockito-all")
+      "junit" % "junit" % "4.12" % "test"
     ),
     resolvers += "twitter-repo" at "http://maven.twttr.com",
 
@@ -183,6 +183,7 @@ object Scrooge extends Build {
       "org.apache.thrift" % "libthrift" % "0.5.0-1",
       "com.github.scopt" %% "scopt" % "3.2.0",
       "com.novocode" % "junit-interface" % "0.8" % "test->default" exclude("org.mockito", "mockito-all"),
+      "com.github.spullara.mustache.java" % "compiler" % "0.8.17",
       "org.codehaus.plexus" % "plexus-utils" % "1.5.4",
       "org.slf4j" % "slf4j-log4j12" % "1.6.6" % "test", // used in thrift transports
       "com.google.code.findbugs" % "jsr305" % "1.3.9",
@@ -190,19 +191,6 @@ object Scrooge extends Build {
       finagle("core") exclude("org.mockito", "mockito-all"),
       finagle("thrift") % "test"
     ),
-    libraryDependencies <++= scalaVersion({
-      case version if version.startsWith("2.10.") =>
-        Seq(
-          "com.github.spullara.mustache.java" % "scala-extensions-2.10" % "0.8.17",
-          "com.github.spullara.mustache.java" % "compiler" % "0.8.17"
-        )
-      case version if version.startsWith("2.11.") =>
-        Seq(
-          "com.github.spullara.mustache.java" % "scala-extensions-2.11" % "0.9.0",
-          "com.github.spullara.mustache.java" % "compiler" % "0.9.0"
-        )
-
-    }),
     test in assembly := {},  // Skip tests when running assembly.
     mainClass in assembly := Some("com.twitter.scrooge.Main")
   ).dependsOn(scroogeRuntime % "test")
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -63,7 +63,6 @@ object Scrooge extends Build {
   val sharedSettings = Seq(
     version := libVersion,
     organization := "com.twitter",
-    // 2.11-ification needs more work with mustache
     crossScalaVersions := Seq("2.10.5" , "2.11.7"),
     scalaVersion := "2.10.5",
 
--- a/scrooge-generator/src/main/scala/com/twitter/scrooge/mustache/ScalaObjectHandler.scala
+++ b/scrooge-generator/src/main/scala/com/twitter/scrooge/mustache/ScalaObjectHandler.scala
@@ -1,0 +1,72 @@
+package com.twitter.scrooge.mustache
+
+import collection.JavaConversions._
+import com.github.mustachejava.Iteration
+import com.github.mustachejava.reflect.ReflectionObjectHandler
+import java.io.Writer
+import java.lang.reflect.{Field, Method}
+import runtime.BoxedUnit
+import scala.reflect.ClassTag
+
+/**
+ * This class is borrowed from Mustache.java in scala-extensions. Its package
+ * has been renamed from com.twitter.mustache to com.twitter.scrooge.mustache,
+ * so it won't collide if a user also has included the scala-extensions jar in
+ * their classpath.
+ */
+private[scrooge] class ScalaObjectHandler extends ReflectionObjectHandler {
+
+  // Allow any method or field
+  override def checkMethod(member: Method) {}
+
+  override def checkField(member: Field) {}
+
+  override def coerce(value: AnyRef) = {
+    value match {
+      case m: collection.Map[_, _] => mapAsJavaMap(m)
+      case u: BoxedUnit => null
+      case Some(some: AnyRef) => coerce(some)
+      case None => null
+      case _ => value
+    }
+  }
+
+  override def iterate(iteration: Iteration, writer: Writer, value: AnyRef, scopes: Array[AnyRef]) = {
+    value match {
+      case TraversableAnyRef(t) => {
+        var newWriter = writer
+        t foreach {
+          next =>
+            newWriter = iteration.next(newWriter, coerce(next), scopes)
+        }
+        newWriter
+      }
+      case n: Number => if (n.intValue() == 0) writer else iteration.next(writer, coerce(value), scopes)
+      case _ => super.iterate(iteration, writer, value, scopes)
+    }
+  }
+
+  override def falsey(iteration: Iteration, writer: Writer, value: AnyRef, scopes: Array[AnyRef]) = {
+    value match {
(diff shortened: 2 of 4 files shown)


scrooge-generator/src/main/scala/com/twitter/mustache/ScalaObjectHandler.scala
```diff
@@ -0,0 +1,69 @@
+package com.twitter.mustache
```
eirslett: This class was copied directly from Mustache.java because it wasn't released for Scala 2.11 on Java 7.

adriancole:
sbt tests pass for me locally!

project/Build.scala
```diff
@@ -13,14 +13,11 @@ import scoverage.ScoverageSbtPlugin
 
 object Scrooge extends Build {
   val branch = Process("git" :: "rev-parse" :: "--abbrev-ref" :: "HEAD" :: Nil).!!.trim
-  val suffix = if (branch == "master") "" else "-SNAPSHOT"
```
mosesn: please don't turn this off, it's extremely useful for making changes in libs that depend on other libs that haven't been published yet.

version.sbt
```diff
@@ -0,0 +1 @@
+version in ThisBuild := "3.20.0-SNAPSHOT"
```
mosesn: what does this do?

scrooge-generator/src/main/scala/com/twitter/mustache/ScalaObjectHandler.scala
mosesn: I'm worried that this will cause big problems for folks who depend both on scala-extensions and scrooge.

Can you test that it doesn't cause problems?  Also, we shouldn't expose this publicly–although scrooge-generator isn't technically a library, I'm worried that folks will use it anyway and it'll be painful to remove it after we move to java 8.

eirslett: How about we shade it in? We change the package name and make it package-protected, so even if somebody uses scala-extensions, they won't collide?
`package com.twitter.scrooge.mustache.shaded.ScalaObjectHandler`

project/Build.scala
eirslett: Sorry, I'll put it back in!
Maybe there should be a "snapshot" mode in the build, like
`sbt snapshot compile`? Then it wouldn't be tied to the git branch name. E.g. in a setup with multiple remotes, "eirslett-master" is my local name for the eirslett/master branch, while "twitter-master" is my local name for the twitter/master branch.

version.sbt
eirslett: Mostly it was to move the version declaration its own place since that helps a lot with regex-style code analysis (https://github.com/sbt/sbt-release), but I guess you don't use sbt-release anyways, maybe just sbt publish?
I can move it back to the Build.scala file.

eirslett:
Ok, I changed it back now. Also changed the package name of scala-extensions so it doesn't collide.

scrooge-generator/src/main/scala/com/twitter/mustache/ScalaObjectHandler.scala
adriancole: sorry if I'm dense.. is this class only used by scrooge generator? Not sure how a class with a different name that's used by a generator would conflict with something.

adriancole: oh disregard.. I guess the original version shadowed a file.

adriancole:
:shipit: non-binding

scrooge-generator/src/main/scala/com/twitter/scrooge/mustache/ScalaObjectHandler.scala
```diff
@@ -0,0 +1,72 @@
...
+import runtime.BoxedUnit
+import scala.reflect.ClassTag
+
+/**
+ * This class is borrowed from Mustache.java in scala-extensions. Its package
+ * has been renamed from com.twitter.mustache to com.twitter.scrooge.mustache,
+ * so it won't collide id a user also has included the scala-extensions jar in
+ * their classpath.
+ */
+class ScalaObjectHandler extends ReflectionObjectHandler {
```
mosesn: can we still make this private-y?

version.sbt
mosesn: nah, I just didn't know what it did, I'd be happy if you stuck that into the patch–we'll have to change a small amount of tooling, but you're right that it will make our lives easier.  if you don't feel like putting it back in, not a big deal though.

project/Build.scala
mosesn: I'm not tied to a specific solution.  We can do that, but we should document it clearly and we should do it in util, finagle, ostrich, scrooge, and twitter-server at pretty much the same time so that it doesn't end up piecemeal.

Our feeling was that it would be useful to be always be able to build off of the `master` branch, and that folks who had to change the util / finagle / scrooge / ostrich / twitter-server `develop` branch would probably want it to be able to compile if you had published local snapshot versions of the other projects, so we made it depend on the `-SNAPSHOT` version and included instructions in the CONTRIBUTING.md file (looks like we forgot to make a CONTRIBUTING.md for scrooge though).  Our concern is that we'll sometimes add a new feature in util and then immediately depend on it in a different library, which works without problems in a monorepo, but has problems in a multi-repo scenario.

Maybe the right solution is to make a dbuild for twitter online-service repositories?  I haven't looked super deeply into dbuild.

project/Build.scala
```diff
@@ -93,7 +93,7 @@ object Scrooge extends Build {
     libraryDependencies ++= Seq(
       "org.scalatest" %% "scalatest" % "2.2.4" % "test",
       "org.scalacheck" %% "scalacheck" % "1.12.2" % "test",
-      "junit" % "junit" % "4.10" % "test" exclude("org.mockito", "mockito-all")
+      "junit" % "junit" % "4.12" % "test"
```
mosesn: do we need this, or was this just a cleanliness thing?

mosesn:
:shipit:, just a couple nits.

project/Build.scala
eirslett: Junit doesn't ship with mockito, so we don't ned to exclude it.
I bumped it to 4.12 just to stay up-to-date (version 4.10 is from 29-Sep-2011, version 4.12 is from 04-Dec-2014)
I'm not sure if this builds with 4.10 or not, I didn't really test. Do you want me to change it back to 4.10?

eirslett:
I rebased and pushed. The only change was to make ScalaObjectHandler package-protected.

scrooge-generator/src/main/scala/com/twitter/scrooge/mustache/ScalaObjectHandler.scala
```diff
@@ -0,0 +1,72 @@
...
+import com.github.mustachejava.reflect.ReflectionObjectHandler
+import java.io.Writer
+import java.lang.reflect.{Field, Method}
+import runtime.BoxedUnit
+import scala.reflect.ClassTag
+
+/**
+ * This class is borrowed from Mustache.java in scala-extensions. Its package
+ * has been renamed from com.twitter.mustache to com.twitter.scrooge.mustache,
+ * so it won't collide id a user also has included the scala-extensions jar in
```
nshkrob: Typo: if a user...

project/Build.scala
```diff
@@ -190,19 +191,6 @@ object Scrooge extends Build {
       finagle("core") exclude("org.mockito", "mockito-all"),
       finagle("thrift") % "test"
     ),
-    libraryDependencies <++= scalaVersion({
```
nshkrob: So the new version works because mustache.java 0.8.17 works for scala 2.11 (except the extensions), right?

For future reference, what's not working with the old approach?

nshkrob:
LGTM overall, except the typo and a question.

scrooge-generator/src/main/scala/com/twitter/scrooge/mustache/ScalaObjectHandler.scala
eirslett: good catch; I fixed it.

project/Build.scala
eirslett: There are different versions of things at play here:

- Java 7 or 8
- Scala 2.10 or 2.11
- Mustache.java 0.8.17 or 0.9

We need to support Scala 2.11 with Java 7. Mustache.java from 0.9 onwards only supports Java 8, so we need version 0.8.17 of Mustache.java. Scala-extensions has been published for Scala 2.10, but for Scala 2.11 the only scala-extensions published are for Mustache.java 0.9 (and therefore Java 8). That is the reason why ScalaObjectHandler is inlined - we need ScalaObjectHandler for Java 7 and Scala 2.11.

The problem with the Scrooge build was that it pattern-matched `scalaVersion` and set library versions accordingly, but when building for Scala 2.11 it would only work with Java 8.
Once Scrooge is declared Java-8-only, it can use Mustache.java 0.9 and Scala-extensions 2.11.

project/Build.scala
mosesn: it's probably fine–we use 4.10 internally, but I don't think it'll cause any problems.  we probably can't use 4.12 features yet though.

mosesn:
:shipit: thanks for this, it looks great!

project/Build.scala
```diff
@@ -64,7 +64,7 @@ object Scrooge extends Build {
     version := libVersion,
     organization := "com.twitter",
     // 2.11-ification needs more work with mustache
```
kevinoliver: does this comment still apply? if so, can you add more color? otherwise, please rm.

eirslett: No, this PR fixes it. I removed the comment now.

vkostyukov:
I'm pulling this for the internal review. Will close this PR once it's merged internally.

vkostyukov:
This has been merged internally. Thanks a lot @eirslett! It's a great contribution!

adriancole:
this is great. since this is a big deal, any chance we can get a scrooge release pushed so we can use it from zipkin?

vkostyukov:
@adriancole We've been planning to realease the whole CSL stack (Finagle, Util, Scrooge) on monday next week? Does it work for you?

benjumanji:
As in Monday the 27th? I am marking this day in my calendar \o/

adriancole:
sure. monday sounds great!

eirslett:
Will there be a ceremony? :balloon: :cocktail:

benjumanji:
I don't see any scala 2.11 bits, are they published?

vkostyukov:
@benjumanji they should be: https://repo1.maven.org/maven2/com/twitter/scrooge-core_2.11/3.20.0/. I noticed that search.maven.org shows pretty outdated results.

benjumanji:
@vkostyukov Ahhh. Fool me for relying on search.maven.org. Thanks!

PR #201, opened 2015-08-09
Generated Java code should read/write enums correctly

eirslett:
This is the same as #139 but opened against the develop branch. It's one of the commits that is needed in order to support Java code generation properly (a few more PRs coming soon)


mosesn:
LGTM, sorry it took so long to review!

mosesn:
Actually, it looks like this PR is for experimental-java, which doesn't exist anymore.  Sorry about that!

PR #200, opened 2015-08-05
rebase experimental-java RHS struct PR

pavanch007:
rebased the pull request from @eirslett  https://github.com/twitter/scrooge/pull/137

PR #202, opened 2015-08-09
scrooge: Add explicit type arguments when generating Java code

eirslett:
Problem

The Java compiler cannot infer type arguments (like Scala does) when
we inline function calls. Therefore, we need to call Utilities.makeXXX
with explicit types, e.g. Utilities.<String,String>makeMap().
The problem will only come up when we generate RHS structs
with builders (PR #200).

Example:

```
struct Foo {
  1: list<i32> bars;
}

new Foo.Builder().bars(Utilities.makeList()).build();
```

will fail, while

```
new Foo.Builder().bars(Utilities.<Integer>makeList()).build();
```

will compile.

Solution

Add type arguments to code generation for Utilities.

### DIFF
--- a/scrooge-generator/src/main/scala/com/twitter/scrooge/backend/JavaGenerator.scala
+++ b/scrooge-generator/src/main/scala/com/twitter/scrooge/backend/JavaGenerator.scala
@@ -76,11 +76,15 @@ class JavaGenerator(
     fieldType: Option[FieldType] = None
   ): CodeFragment = {
     val listElemType = fieldType.map(_.asInstanceOf[ListType].eltType)
+    val typeArgument = listElemType match {
+      case Some(elemType) => "<" + genType(elemType).toData + ">"
+      case _ => ""
+    }
     val code =
       list.elems.map { e =>
         genConstant(e, listElemType).toData
       }.mkString(", ")
-    v(s"Utilities.makeList($code)")
+    v(s"Utilities.${typeArgument}makeList($code)")
   }
 
   def genSet(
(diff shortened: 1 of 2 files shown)


eirslett:
I added a test case. The test case needs #200 to be merged before it works.

nshkrob:
Hey, I've tried this out (#200 and this patch). I don't see the typed makeList calls in the output, so I don't think the typeArgument part is triggered. Can you check this please?

eirslett:
You're right - in scrooge-generator/target/scala-2.10/src_managed/test/com/twitter/scrooge/integration_java/Constants.java it's triggering type arguments in simple cases, but not in all the cases... I'll investigate!

eirslett:
There we go - it should be fixed now. I had to cherry-pick @pavanch007's work into this PR, because both commits were needed for the fix.

nshkrob:
Ok, now I see the typed calls. Input:

struct Foo {
  1: list<i32> bars;
}

const Foo f = {"bars": [3,4] }

In Constants.java:

public static final Foo f = new Foo.Builder().bars(Utilities.<Integer>makeList(3, 4)).build();

But the java output also compiles for me without the <Integer>. Is there an example where the old code fails? Anything else I'm missing?

eirslett:
I think GitHub removed the `<>` angle brackets from your code :confused:

The empty list (or empty map) is the case where it doesn't compile:

```new Foo.Builder().bars(Utilities.makeList()).build();```

But it will build with

```new Foo.Builder().bars(Utilities.<Integer>makeList()).build();```

nshkrob:
LGTM! Pulling this in.

PR #204, opened 2015-09-04
Allow eponymously named services.

aerskine:
Fixes the import clash with the generated code a thrift service named `Service` and the import of  `com.twitter.finagle.Service`. 

(Conceivably, the same problem could occur if someone created a thrift service named `TReusableMemoryTransport` or any of the other imported names in `finagleService.scala` .. but luckily I don't work somewhere where people would be so awkward :) )

### DIFF
--- a/scrooge-generator/src/main/resources/javagen/finagleService.java
+++ b/scrooge-generator/src/main/resources/javagen/finagleService.java
@@ -1,6 +1,5 @@
 package {{package}};
 
-import com.twitter.finagle.Service;
 import com.twitter.finagle.SourcedException;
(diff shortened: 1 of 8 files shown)


nshkrob:
LGTM, thanks! I'll pull it in.

aerskine:
@nshkrob Thanks! Anything else I need to do?

nshkrob:
@aerskine No, this has been merged at https://github.com/twitter/scrooge/commit/028be5ff57005a454520c59a40e68624838601c4.
We'll probably release a new scrooge version with this patch in a few days.

PR #207, opened 2015-10-05
Add twitter resolver to plugins.sbt

gidsg:
You need the twitter resolver in order to get the latest dependencies.

Without this line, adding the latest version of the plugin fails with the error:

```
[warn] 	::::::::::::::::::::::::::::::::::::::::::::::
[warn] 	::          UNRESOLVED DEPENDENCIES         ::
[warn] 	::::::::::::::::::::::::::::::::::::::::::::::
[warn] 	:: org.apache.thrift#libthrift;0.5.0-1: not found
[warn] 	::::::::::::::::::::::::::::::::::::::::::::::
```

### DIFF
--- a/SBTPlugin.html
+++ b/SBTPlugin.html
@@ -60,6 +60,9 @@
 <h1>SBT Plugin<a class="headerlink" href="#sbt-plugin" title="Permalink to this headline">¶</a></h1>
 <p>Add a line like this to your <cite>project/plugins.sbt</cite> file:</p>
 <div class="highlight-text"><div class="highlight"><pre>addSbtPlugin(&quot;com.twitter&quot; %% &quot;scrooge-sbt-plugin&quot; % &quot;3.18.1&quot;)
+
+resolvers += &quot;twitter-repo&quot; at &quot;https://maven.twttr.com&quot;
+
 </pre></div>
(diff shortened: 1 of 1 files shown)


SBTPlugin.html
```diff
@@ -60,6 +60,9 @@
 <h1>SBT Plugin<a class="headerlink" href="#sbt-plugin" title="Permalink to this headline">¶</a></h1>
 <p>Add a line like this to your <cite>project/plugins.sbt</cite> file:</p>
```
mosesn: could you change the copy here to reflect that there's more than one line you need to add?

mosesn:
LGTM, except for the "one line" thing

nshkrob:
Thanks for improving the docs!
The html file in the branch gh-pages is generated from the file SBTPlugin.rst. Please edit that one directly and then we'll push the updated version of the docs. 
If you want to verify your changes you can generate html from the rst files by running `./sbt scrooge-doc/make-site` (you need sphinx-build for that).

gidsg:
OK thanks for your feedback @nshkrob @mosesn. I have made the suggested changes in #208 as this was branched incorrectly off of `gh-pages`.

PR #210, opened 2015-10-17
Fix name conflict in ScroogeSBT

### LINKED ISSUE #209: "compile" function in ScroogeSBT conflicts with sbt.Keys.compile
adding the scrooge sbt plugin causes sbt to import all of it's keys into the namespace of a `built.sbt` file. so in `build.sbt`, the top of the file magically receives:

```scala
import _root_.com.twitter.scrooge.ScroogeSBT._
import Keys._
```
this includes the `compile` function here: https://github.com/twitter/scrooge/blob/develop/scrooge-sbt-plugin/src/main/scala/com/twitter/ScroogeSBT.scala#L13 , but it collides with https://github.com/sbt/sbt/blob/1.0.x/main/src/main/scala/sbt/Keys.scala#L188

this becomes a problem when you want to have target specific settings, for example:

```scala
// in build.sbt
wartremoverErrors in (Compile, compile) ++= Seq(...)

// compile error:
//
// .../build.sbt:43: error: reference to compile is ambiguous;
// it is imported twice in the same scope by
// import _root_.com.twitter.scrooge.ScroogeSBT._
// and import Keys._
//  wartremoverErrors in (Compile, compile) ++= Seq(...)
```
as a temporary workaround, you can do:

```scala
// at the top of build.sbt
import Keys.{compile => sbtCompile}
...
wartremoverErrors in (Compile, sbtCompile) ++= Seq(...)
```
renaming the function in scrooge should fix it.

matthandlersux:
@mosesn fixes #209 

the `compile` method is public however, so it's unclear if anyone in the wild references this directly in their build.sbt. if so, this would be considered a breaking change. 

an alternative solution would be to just make the method private. i don't know enough about `scrooge` yet to tell whether this is preferred, please let me know.

note that i updated the README to reflect the extra step required for me to build this package. also note that in order to get everything to compile, i had to explicitly build against scala 2.10.6. i'm not sure if this is expected.

### DIFF
--- a/scrooge-sbt-plugin/src/main/scala/com/twitter/ScroogeSBT.scala
+++ b/scrooge-sbt-plugin/src/main/scala/com/twitter/ScroogeSBT.scala
@@ -10,7 +10,7 @@ object ScroogeSBT extends AutoPlugin {
   private[this] def generatedExtensionPattern(language: String): String =
     if (language.endsWith("java")) "*.java" else "*.scala"
 
-  def compileThrift(
+  private[this] def compile(
     log: Logger,
     outputDir: File,
     thriftFiles: Set[File],
@@ -238,7 +238,7 @@ object ScroogeSBT extends AutoPlugin {
       // for some reason, sbt sometimes calls us multiple times, often with no source files.
       if (isDirty && sources.nonEmpty) {
         out.log.info("Generating scrooge thrift for %s ...".format(sources.mkString(", ")))
-        compileThrift(out.log, outputDir, sources.toSet, inc.toSet, ns, language, opts.toSet)
+        compile(out.log, outputDir, sources.toSet, inc.toSet, ns, language, opts.toSet)
       }
       (outputDir ** generatedExtensionPattern(language)).get.toSeq
     },
--- a/scrooge-sbt-plugin/src/main/scala/com/twitter/ScroogeSBT.scala
+++ b/scrooge-sbt-plugin/src/main/scala/com/twitter/ScroogeSBT.scala
@@ -10,7 +10,7 @@ object ScroogeSBT extends AutoPlugin {
   private[this] def generatedExtensionPattern(language: String): String =
     if (language.endsWith("java")) "*.java" else "*.scala"
 
-  def compile(
+  def compileThrift(
     log: Logger,
     outputDir: File,
     thriftFiles: Set[File],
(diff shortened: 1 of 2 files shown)


scrooge-sbt-plugin/src/main/scala/com/twitter/ScroogeSBT.scala
```diff
@@ -10,7 +10,7 @@ object ScroogeSBT extends AutoPlugin {
   private[this] def generatedExtensionPattern(language: String): String =
     if (language.endsWith("java")) "*.java" else "*.scala"
 
-  def compile(
+  def compileThrift(
```
mosesn: let's mark it `private[this]`

mosesn:
I think you're right that it's safe to rename it.  From looking at the AutoPlugin docs, it doesn't look like it's exposed anywhere.

Good catch on scrooge-core, that has been a mess for a while, we should get around to fixing that.  You could only build against 2.10.6 as in other versions of 2.10 didn't work, or 2.11 didn't work?  That's somewhat odd, because we recently switched to using 2.11 internally.

Could you add a commit message that's structured as we suggest in finagle's [CONTRIBUTING][0] document?  We unfortunately don't have one for scrooge, we should fix that, sorry!

[0]: https://github.com/twitter/finagle/blob/develop/CONTRIBUTING.md

matthandlersux:
@mosesn updated

i could not compile scrooge against 2.11 after successfully building all of the dependencies. unfortunately, i closed the window, but the specific compile error was something about scala 2.11 not being able to use macros that were built against another version of scala. setting `scalaVersion := "2.10.6"` in all of the dependencies and then rebuilding them fixed it.

mosesn:
Oh, weird.  It sounds like we're pulling in a library from the wrong version somewhere.  We don't have the bandwidth to look into it now, but we're planning on dropping support for 2.10 soon-ish (lol) so we'll definitely get that sorted out before that happens.

mosesn:
LGTM, I'll try to get one more set of eyes on this before we merge it in.  Thanks again!

kevinoliver:
lgtm too. thanks @matthandlersux

matthandlersux:
getinsertpic.com

vkostyukov:
I'm pulling this internally. Will close this PR once it's merged. Thanks @matthandlersux!

vkostyukov:
Just merged this internally. Thanks again @matthandlersux!

PR #211, opened 2015-10-17
Add per function name stats tracking for generated scala service code.

zfy0701:
this is already supported in client side, so I think it make sense to add it in service side.

I didn't add this in generated java code because java service didn't even take a StatsReciever as constructor parameter, so there will be a big change.

### DIFF
--- a/scrooge-generator/src/main/resources/scalagen/finagleService.scala
+++ b/scrooge-generator/src/main/resources/scalagen/finagleService.scala
@@ -3,7 +3,7 @@ package {{package}}
 import com.twitter.finagle.Thrift
 import com.twitter.finagle.stats.{NullStatsReceiver, StatsReceiver}
 import com.twitter.scrooge.{ThriftStruct, TReusableMemoryTransport}
-import com.twitter.util.Future
+import com.twitter.util.{Future, Return, Throw, Throwables}
 import java.nio.ByteBuffer
 import java.util.Arrays
 import org.apache.thrift.protocol._
@@ -21,15 +21,22 @@ class {{ServiceName}}$FinagleService(
   iface: {{ServiceName}}[Future],
   protocolFactory: TProtocolFactory,
   stats: StatsReceiver,
-  maxThriftBufferSize: Int
+  maxThriftBufferSize: Int,
+  serviceName: String
 ) extends {{finagleServiceParent}}{{#hasParent}}(iface, protocolFactory, stats, maxThriftBufferSize){{/hasParent}} {
   import {{ServiceName}}._
 
   def this(
     iface: {{ServiceName}}[Future],
+    protocolFactory: TProtocolFactory,
+    stats: StatsReceiver,
+    maxThriftBufferSize: Int
+  ) = this(iface, protocolFactory, stats, maxThriftBufferSize, "{{ServiceName}}")
+
+  def this(
+    iface: {{ServiceName}}[Future],
     protocolFactory: TProtocolFactory
   ) = this(iface, protocolFactory, NullStatsReceiver, Thrift.maxThriftBufferSize)
-
 {{^hasParent}}
 
   private[this] val tlReusableBuffer = new ThreadLocal[TReusableMemoryTransport] {
@@ -119,6 +126,7 @@ class {{ServiceName}}$FinagleService(
   // ---- end boilerplate.
 
 {{/hasParent}}
+  private[this] val scopedStats = if (serviceName != "") stats.scope(serviceName) else stats
 {{#functions}}
   {{>function}}
 {{/function}}
--- a/scrooge-generator/src/main/resources/scalagen/finagleServiceFunction.scala
+++ b/scrooge-generator/src/main/resources/scalagen/finagleServiceFunction.scala
@@ -1,20 +1,33 @@
+private[this] object {{__stats_name}} {
+  val RequestsCounter = scopedStats.scope("{{serviceFuncNameForWire}}").counter("requests")
+  val SuccessCounter = scopedStats.scope("{{serviceFuncNameForWire}}").counter("success")
+  val FailuresCounter = scopedStats.scope("{{serviceFuncNameForWire}}").counter("failures")
+  val FailuresScope = scopedStats.scope("{{serviceFuncNameForWire}}").scope("failures")
+}
 addFunction("{{serviceFuncNameForWire}}", { (iprot: TProtocol, seqid: Int) =>
   try {
+    {{__stats_name}}.RequestsCounter.incr()
     val args = {{funcObjectName}}.Args.decode(iprot)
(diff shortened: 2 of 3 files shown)


scrooge-generator/src/main/resources/scalagen/finagleService.scala
```diff
@@ -21,7 +21,8 @@ class {{ServiceName}}$FinagleService(
   iface: {{ServiceName}}[Future],
   protocolFactory: TProtocolFactory,
   stats: StatsReceiver,
-  maxThriftBufferSize: Int
+  maxThriftBufferSize: Int,
+  serviceName: String = "{{ServiceName}}"
```
mosesn: hmm, instead of using a default argument, could you overload the constructor?  I think that will ease binary compatibility.

scrooge-generator/src/main/resources/scalagen/finagleService.scala
```diff
@@ -119,6 +120,7 @@ class {{ServiceName}}$FinagleService(
   // ---- end boilerplate.
 
 {{/hasParent}}
+  private[this] val scopedStats = if (serviceName != "") stats.scope(serviceName) else stats
```
mosesn: is it valuable to scope these stats with a name?  it seems unlikely that a single service will expose more than one thrift server.

scrooge-generator/src/main/resources/scalagen/finagleServiceFunction.scala
```diff
@@ -15,6 +22,12 @@ addFunction("{{serviceFuncNameForWire}}", { (iprot: TProtocol, seqid: Int) =>
       }
 {{/exceptions}}
       case e => Future.exception(e)
+    } respond {
```
mosesn: let's use the style `}.respond {` here.

mosesn:
How have you tested this?  Do you know how easy it would be to add a test for it?  If we decide to go forward with asking for a serviceName parameter, we'll need to add that to the finagle-thrift rich construction methods too.

LGTM in general, could you add a commit message as suggested in finagle's [CONTRIBUTING][0] section?  Unfortunately, scrooge doesn't have one yet, we should get on fixing that!

[0]: https://github.com/twitter/finagle/blob/develop/CONTRIBUTING.md

scrooge-generator/src/main/resources/scalagen/finagleServiceFunction.scala
zfy0701: the generated code looks like follow:
it would be better to keep '`} respond { `, or we need to change all styles accorrdingly
```
try {
      __stats_call.RequestsCounter.incr()
      val args = Call.Args.decode(iprot)
      iprot.readMessageEnd()
      (try {
        iface.call(args.name)
      } catch {
        case e: Exception => Future.exception(e)
      }) flatMap { value: Seq[Int] =>
        reply("call", seqid, Call.Result(success = Some(value)))
      } rescue {
        case e => Future.exception(e)
      } respond {
        case Return(_) =>
          __stats_call.SuccessCounter.incr()
        case Throw(ex) =>
          __stats_call.FailuresCounter.incr()
          __stats_call.FailuresScope.counter(Throwables.mkString(ex): _*).incr()
      ```

scrooge-generator/src/main/resources/scalagen/finagleService.scala
zfy0701: currently, all the server related stats are scoped with name. following are examples from my server, so I think we just follow the convention
"srv/XXServer/closechans": 84,
"srv/XXServer/closed": 84,
"srv/XXServer/closes": 0,

scrooge-generator/src/main/resources/scalagen/finagleService.scala
zfy0701: done

scrooge-generator/src/main/resources/scalagen/finagleServiceFunction.scala
mosesn: we've been switching over lazily–if you could do a bunch, it would be great, but it's ok to just switch over lazily too.

scrooge-generator/src/main/resources/scalagen/finagleService.scala
mosesn: ok, can you open a PR on finagle-thrift too then, so that we can thread the name through properly with the rich client interface?

zfy0701:
add overload constructor, add description.

yes I did test and I verified the stats in my server compiled from the newly generated code.

One interesting thing, it seems if you are using new CSL style generated, finagle will generated those stats for you:

https://github.com/twitter/finagle/blob/develop/finagle-thrift/src/main/scala/com/twitter/finagle/thrift/ThriftServiceIface.scala#L140

my change doesn't conflict with it because the new style will not use the code generated here.

scrooge-generator/src/main/resources/scalagen/finagleServiceFunction.scala
zfy0701: done

scrooge-generator/src/main/resources/scalagen/finagleService.scala
zfy0701: will do

zfy0701:
also please let me know if you want me to create PR on develop branch instead of master

mosesn:
I *think* it should be fine, I don't think there are any relevant conflicts between develop and master right now.

Re: ServiceIface–this is still useful because we only have ServiceIface support for clients, afaik.

zfy0701:
PR for finagle-thrift created:
https://github.com/twitter/finagle/pull/431

is there any remaining issue here?

mosesn:
LGTM, I'll try to get more eyes on it :+1:

blackicewei:
LGTM. I'll pull it in internally.

vkostyukov:
This has been merged internally. Thanks @zfy0701!

zfy0701:
just wondering when will this change be merged into OSS repo, it has already been taking a very long time.
let me know if there is anything wrong with the change.

ryanoneill:
Hi @zfy0701, this was merged in internally and pushed back out in this commit (https://github.com/twitter/scrooge/commit/fb4ab4750eb37a337754fc60c196cd6b4b740264). Unfortunately, it looks like the metadata was lost on this pull request. Our apologies for any confusion.

mosesn:
ahhh, this is a bummer.  sorry @zfy0701, our tooling around making sure that authorship is retained is really bad.  I've filed a ticket to try to fix this, and I'll talk to some people to try to get the ball rolling so this doesn't happen again.  again, we're really sorry that this happened.

zfy0701:
Np at all. Nothing make me more happy to see the actual code went in :)

PR #214, opened 2015-11-25
Fix nullary warning within `UnknownUnionField`.

nathankleyn:
The `containedValue` field returns `Unit` but does not follow the standard of having parenthesis after the method name (`containedValue()`). Thus, a warning is produced:

```scala
[warn] foo.scala:240: side-effecting nullary methods are discouraged: suggest defining as `def containedValue()` instead
[warn]     def containedValue: Unit = ()
[warn]         ^
```

This fix simply adds the parenthesis to the method name.

### DIFF
--- a/scrooge-generator/src/main/resources/scalagen/union.scala
+++ b/scrooge-generator/src/main/resources/scalagen/union.scala
@@ -219,7 +219,7 @@ object {{StructName}} extends ThriftStructCodec3[{{StructName}}] {
       private val field: TFieldBlob)
     extends {{StructName}} {
 
-    def containedValue: Unit = ()
+    def containedValue(): Unit = ()
 
     def unionStructFieldInfo: _root_.scala.Option[ThriftStructFieldInfo] = _root_.scala.None
 


kevinoliver:
Cool. Since its still callable without the `()s` I guess its not really a breaking API change.

That said, mind updating the one usage of it in `ThriftStructMetaDataSpec.scala`?

nathankleyn:
@kevinoliver Good catch! I've added the parenthesis to that call in `ThriftStructMetaDataSpec`.

nathankleyn:
Builds are failing, made that last edit away from my computer - will look at this ASAP tonight.

scrooge-generator/src/test/scala/com/twitter/scrooge/ThriftStructMetaDataSpec.scala
```diff
@@ -162,7 +162,7 @@ class ThriftStructMetaDataSpec extends Spec {
     val bonk = MorePerfectUnion.Bonk(Bonk("message", 55))
     assert(Some(bonkFieldInfo) == bonk.unionStructFieldInfo)
 
-    bonk.containedValue match {
+    bonk.containedValue() match {
```
taylorleese: This fails to compile because ```containedValue``` in ```MorePerfectUnion``` is parameterless and it's return type isn't actually ```Unit``` so this shouldn't change.

nathankleyn: Quite correct - have fixed now and is ready for checking as soon as the builds are done.

taylorleese:
This was actually fixed as part of https://github.com/twitter/scrooge/commit/10d0b51c800d2fb61f93551d9fb17f401b823963#diff-edb9a11655d6791ed6e9db2a8b2ec521.

PR #215, opened 2015-12-08
Use explicit `java.lang.Override` annotation.

cab:
Problem

Using just `@Override` has a potential for naming conflicts (e.g. `struct Override`).

Solution

Generated code uses `@java.lang.Override` instead.

### DIFF
--- a/scrooge-generator/src/main/resources/androidgen/struct_inner.mustache
+++ b/scrooge-generator/src/main/resources/androidgen/struct_inner.mustache
@@ -85,7 +85,7 @@ public {{#is_final}}final {{/is_final}}{{#in_class}}static {{/in_class}}class {{
     return new {{name}}(this);
   }
 
-  @Override
+  @java.lang.Override
   public void clear() {
     {{#fields}}
     {{#consolidate_newlines}}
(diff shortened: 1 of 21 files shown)


scrooge-generator/src/main/resources/androidgen/struct_inner.mustache
```diff
@@ -85,7 +85,7 @@ public {{#is_final}}final {{/is_final}}{{#in_class}}static {{/in_class}}class {{
     return new {{name}}(this);
   }
 
-  @Override
+  @java.lang.Override
```
mosesn: since we're trying to avoid collisions, let's use _root_.java.lang.Override for good measure.

mosesn:
LGTM except for one thing!

scrooge-generator/src/main/resources/androidgen/struct_inner.mustache
cab: Sorry if I'm misunderstanding (or wrong :)), but isn't `_root_` only supported in Scala?

mosesn: ah, indeed!  my bad.

mosesn:
LGTM

kevinoliver:
lgtm too, thanks @cab

taylorleese:
@kevinoliver Since this is in the scrooge generator it would be good to get this merged before we release a new version of the scrooge plugin this week.

kevinoliver:
I'll get to it this week.

kevinoliver:
@cab just got this merged in locally. should appear on the `develop` branch this monday. sorry for the delay. thanks again for the patch.

PR #224, opened 2016-03-11
Add abstract value, name, and originalName to sealed enum trait

shshe:
Problem

For an enum, generated case objects have a `value`, `name` and `originalName`
field, but these are inaccessible when working with the enum's sealed trait.

Solution

This adds these fields to the sealed trait so we can access them when working
with objects typed as the sealed trait.


mosesn:
LGTM, thanks for the contribution!

kevinoliver:
During the review internally, someone pointed out that `value` and `name` already exist on `c.t.scrooge.ThriftEnum`.  Can you instead add `originalName` there? This would be a breaking API change so it should be documented as such in `scrooge/CHANGES` but I think its worth it.

kevinoliver:
Actually, right after I hit "Comment", I realized the default implementation could be to return `name` and thus it would not break the API.

shshe:
Thanks for the feedback! Closing this PR and opened up #227.

PR #226, opened 2016-03-17
Add originalName to ThriftEnum

shshe:
Problem

For an enum, generated case objects have an abstract name and value field, but not originalName.

Solution

Add originalName to the ThriftEnum trait and generated EnumUnknown case classes.

See #224 for prior PR.


shshe:
Thanks for the feedback! I didn't see that name and value were already in ThriftEnum :) Closing this PR and opened #226.

PR #227, opened 2016-03-17
Add originalName to ThriftEnum

shshe:
Problem

For an enum, generated case objects have an abstract name and value field, but not originalName.

Solution

Add originalName to the ThriftEnum trait and generated EnumUnknown case classes. Add default implementation of originalName to return name to avoid breaking the API.

See #224 for prior PR.

### DIFF
--- a/scrooge-core/src/main/scala/com/twitter/scrooge/ThriftEnum.scala
+++ b/scrooge-core/src/main/scala/com/twitter/scrooge/ThriftEnum.scala
@@ -5,6 +5,6 @@ import org.apache.thrift.TEnum
 trait ThriftEnum extends TEnum {
   def value: Int
   def name: String
-  def originalName: String = name
+  def originalName: String
   def getValue = value
 }
--- a/scrooge-core/src/main/scala/com/twitter/scrooge/ThriftEnum.scala
+++ b/scrooge-core/src/main/scala/com/twitter/scrooge/ThriftEnum.scala
@@ -5,5 +5,6 @@ import org.apache.thrift.TEnum
 trait ThriftEnum extends TEnum {
   def value: Int
   def name: String
+  def originalName: String = name
   def getValue = value
 }
--- a/scrooge-core/src/main/scala/com/twitter/scrooge/ThriftEnum.scala
+++ b/scrooge-core/src/main/scala/com/twitter/scrooge/ThriftEnum.scala
@@ -2,9 +2,15 @@ package com.twitter.scrooge
 
 import org.apache.thrift.TEnum
 
+
 trait ThriftEnum extends TEnum {
   def value: Int
   def name: String
+
+  /**
+    * The original name for the enum value as defined in the input Thrift IDL file.
+    */
   def originalName: String
+
   def getValue = value
 }
--- a/scrooge-generator-tests/src/test/resources/gold_file_output_scala/com/twitter/scrooge/test/gold/thriftscala/RequestType.scala
+++ b/scrooge-generator-tests/src/test/resources/gold_file_output_scala/com/twitter/scrooge/test/gold/thriftscala/RequestType.scala
@@ -15,7 +15,7 @@ case object RequestType {
   case object Create extends com.twitter.scrooge.test.gold.thriftscala.RequestType {
     val value = 1
     val name = "Create"
-    override val originalName = "Create"
+    val originalName = "Create"
   }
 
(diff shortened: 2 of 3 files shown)


shshe:
@kevinoliver Used your suggestion to have the default implementation of `originalName` to return `name`, but this requires an `override val originalName = ...` on the implementing case classes. Doesn't this still break the API?

mosesn:
@shshe no, overrides are binary compatible.  https://docs.oracle.com/javase/specs/jls/se7/html/jls-13.html

shshe:
Good to know - thanks!

mosesn:
Adding a new method to an interface is not binary compatible, but I think it's OK.  What do you think of not providing an implementation in ThriftEnum, and requiring implementors of ThirftEnum to provide an implementation/

shshe:
@mosesn: Removed the default implementation of `originalName` from `ThriftEnum`. Previously generated EnumUnknown* case class definitions will have to be regenerated with this new template to include the `originalName` field.

mosesn:
:+1: I'm happy with this.  @kevinoliver thoughts?

scrooge-generator/src/main/resources/scalagen/enum.scala
```diff
@@ -18,6 +18,7 @@ case object {{EnumName}} {
 
   case class EnumUnknown{{EnumName}}(value: Int) extends {{package}}.{{EnumName}} {
     val name = "EnumUnknown{{EnumName}}" + value
+    val originalName = name
```
kevinoliver: 2 small things:
# make this a `def` to keep the object size smaller
# include the return type `String` since this is public (mind doing it on `name` above while you are here?)

scrooge-generator-tests/src/test/resources/gold_file_output_scala/com/twitter/scrooge/test/gold/thriftscala/RequestType.scala
```diff
@@ -30,6 +30,7 @@ case object RequestType {
 
   case class EnumUnknownRequestType(value: Int) extends com.twitter.scrooge.test.gold.thriftscala.RequestType {
     val name = "EnumUnknownRequestType" + value
+    val originalName = name
```
kevinoliver: same comments as below

scrooge-core/src/main/scala/com/twitter/scrooge/ThriftEnum.scala
```diff
@@ -5,5 +5,6 @@ import org.apache.thrift.TEnum
 trait ThriftEnum extends TEnum {
   def value: Int
   def name: String
+  def originalName: String
```
kevinoliver: can we scaladoc what `originalName` represents? 

i'm going to forget by the next time i look at this code.

kevinoliver:
just a few small requests from me @shshe

scrooge-generator/src/main/resources/scalagen/enum.scala
```diff
@@ -8,17 +8,17 @@ case object {{EnumName}} {
 {{#values}}
   {{valuedocstring}}
   case object {{name}} extends {{package}}.{{EnumName}} {
-    val value = {{value}}
-    val name = "{{name}}"
-    val originalName = "{{originalName}}"
+    val value: Int = {{value}}
+    val name: String = "{{name}}"
+    val originalName: String = "{{originalName}}"
```
shshe: @kevinoliver Addressed your concerns. Let me know if there's anything else!

mosesn:
@shshe thanks for the comment, we unfortunately don't get an email when you push a new commit :(

kevinoliver:
lgtm, thanks. i'll try to get this pulled in locally.

kevinoliver:
@shshe unfortunately, i'm struggling to get this merged locally — we end up with incompatibilities both with a default implementation and without one. this is mostly an artifact of how we use scrooge-generator internally and isn't due to your change. still talking to folks about it.

kevinoliver:
ok, @mosesn came up with a path forward but i won't be able to get to it for about 2 weeks. sorry for the delays @shshe.

kevinoliver:
@shshe ok first step is in and working on the second step today.

kevinoliver:
Alright! The second half got merged in this morning and should get pushed out to the `develop` branch in the next push (which should be today).

Thanks again for your patience @shshe and the patch.

Issue #229, opened 2016-03-25
Scrooge maven plugin is overly restrictive


fzakaria:
I believe the maven plugin for scrooge is too restrictive, especially for cases where not *all* thrift model dependencies have the `idl` classifier

Here is the relevant code from `AbstractMavenScroogeMojo`

```java
    for (Artifact artifact : deps) {
      // This artifact has an idl classifier.
      if (isIdlCalssifier(artifact, classifier)) {
        thriftDependencies.add(artifact);
      } else {
        if (isDepOfIdlArtifact(artifact, depsMap)) {
          // Fetch idl artifact for dependency of an idl artifact.
          try {
            Artifact idlArtifact = MavenScroogeCompilerUtil.getIdlArtifact(
              artifact,
              artifactFactory,
              artifactResolver,
              localRepository,
              remoteArtifactRepositories,
              classifier);
            thriftDependencies.add(idlArtifact);
          } catch (MojoExecutionException e) {
            /* Do nothing as this artifact is not an idl artifact
             binary jars may have dependency on thrift lib etc.
             */
            getLog().debug("Could not fetch idl jar for " + artifact);
          }
        }
      }
    }
    return thriftDependencies;
  }
```

Specifically consider the use-case where someone has a Maven artfiact `ThriftDemoModel` with the classifier `idl`.
This artifact depends on `finatra-thrift_2.11` accordingly so that it can make use of `finatra-thrift/finatra_thrift_exceptions.thrift`

The code above will hit the `MavenScroogeCompilerUtil` however it will attempt to use the classifier `idl` and in this particular case, finatra-thrift is not published with any classifier.

The solution to have the classifier as empty or null won't work either because it doesn't allow mixing & matching of dependencies that may /may not use the classifier.

###Solution

A better solution would be once the code has determined `isIdlCalssifier`, it should get all *transitive* dependencies and simply add them to `thriftDependencies`

### Additional bug

There is an additional bug with the current solution. When performing `isDepOfIdlArtifact` it uses the `getDependencyTrail` method. The problem is, if a dependency of an IDL package **AND** the current root module, the dependencyTrail will be the shortest path and not even include the IDL dependency.

PR #230, opened 2016-03-27
Maven plugin

fzakaria:
Problem

This PullRequest is in response to [Issue 229](https://github.com/twitter/scrooge/issues/229).
Essentially, I was going through the tutorial for [finatra](https://twitter.github.io/finatra/) with Thrift and had made a new Java project `ThriftModelExample` with classifier `idl` that held only my thrift file.

My thrift file included `"finatra-thrift/finatra_thrift_exceptions.thrift"` from the `finatra-thrift_2.11` artifact; however this artifact does not have the `idl` classifier.

Solution

The original solution was to include the following:
1. A fully configurable list for the Mojo of `Modules` (groupId & artifactId) that are forced included to be searched for thrift files.
2. For any Artifact that has the IDL classifier the plugin should *automatically* include all transitive dependencies of that artifact.  This makes sense for all cases such as described in the problem statement above where you would want to decompose your Thrift files into many modules and make use of dependency resolution.
3. While I was at it, I also removed dead code that was never referenced
4. While I was at it, I migrated the plugin to use the maven plugin Annotations

Result

All dependencies of a project that is found to be either of classifier `idl` or forced inclusion has all of its dependencies included which allows for the proper modularization of thrift files.
It is also less strict about the classifier naming.

Con
The downside to this change is that I had to remove the test cases.
This new code makes use of the Aether maven API, however there does not seem to be a way to retrofit that into the maven plugin test harness.


fzakaria:
I've uploaded - [ThriftExplorer](https://github.com/fzakaria/ThriftExplorer/tree/d43ce81037d36070dd6149f659f04039551d0be5) to demo the problem and solution.

You can see the [pom.xml](https://github.com/fzakaria/ThriftExplorer/blob/d43ce81037d36070dd6149f659f04039551d0be5/ThriftExplorerModel/pom.xml) for the package that contains my `idl` 

It expect the *4.6.0-SNAPSHOT* so please do a maven install to install into your local repository

scrooge-maven-plugin/src/test/java/com/twitter/ScroogeMavenPluginTest.java
```diff
@@ -1,102 +0,0 @@
...
-import org.apache.commons.collections.CollectionUtils;
-import org.apache.maven.plugin.MojoExecutionException;
-import org.apache.maven.plugin.testing.AbstractMojoTestCase;
-import org.codehaus.plexus.util.FileUtils;
-
-import java.io.File;
-import java.util.Arrays;
-import java.util.Collection;
-
-public class ScroogeMavenPluginTest extends AbstractMojoTestCase {
```
mosesn: these tests seem useful, why are you deleting them?

mosesn:
Sorry it took us so long to review this!  I don't know maven well, and we don't use it internally anymore, so our review will unfortunately have to be pretty cursory.  Is there any chance you can replace the tests that don't work anymore with a new test that does something similar?  LGTM otherwise.

fzakaria:
@mosesn It's been a while since I worked on this codebase. I believe those tests were deleted because I couldn't find out how to transition them to the newer maven plugin model (annotation based) - I am also not terribly familiar with Maven plugin development but needed to fix the root issue for myself.

If we can keep the PR open a bit longer I can give it another stab on the weekend of reintroducing the tests.

mosesn:
That would be great!  No rush =)

mosesn:
@fzakaria hey, have you had a chance to take a look at this again?

mosesn:
@fzakaria I'm thinking maybe we should merge it without the tests?  what do you think?

fzakaria:
Let me check out the PR today and I'll go over it before noon PST.

mosesn:
Rad, thanks!

fzakaria:
@mosesn I remerged from develop and re-added the test class.
I had to comment out two of the tests cause I couldn't figure out how to make them work.
(I couldn't find any great documentation with incorporating aether with Maven plugin)

cacoco:
@fzakaria your most recent changes seem to have a merge conflict (scrooge-maven-plugin/pom.xm), would you mind merging latest from the develop branch and trying to fix the conflicts. Also, I think we're going to want a way to have tests that are at least similar to the ones commented out.

bryce-anderson:
@fzakaria, are you still working on this?

fzakaria:
@bryce-anderson I'll close.
I do not have time to continue commiting to this branch.

Hopefully if someone hits same issue they can see the work done.
Thanks!

PR #234, opened 2016-05-07
modify thrift output folder to prevent clashing with other plugins

joprice:
The current default for `scroogeThriftOutputFolder` causes compilation errors when used alongside other sbt plugins that generate source, the most common being sbt-buildinfo. This solves it by generating the code in a subfolder of the generated sources folder. This also adds scripted setup to test the sbt plugin. The test can be run with `scrooge-sbt-plugin/scripted`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,9 +29,16 @@ before_script:
   - ./bin/travisci
   - travis_retry ./sbt ++$TRAVIS_SCALA_VERSION update
 
-script: 
-  - travis_retry ./sbt ++$TRAVIS_SCALA_VERSION clean coverage test coverageReport
-  - ./sbt ++$TRAVIS_SCALA_VERSION coverageAggregate
+
+script: "echo no op"
+
+matrix:
+  include:
+    - scala: 2.10.6
+      # the 'scripted' sbt plugin tests are only run for 2.10
+      script: travis_retry ./sbt ++$TRAVIS_SCALA_VERSION scrooge-sbt-plugin/scripted && ./sbt ++$TRAVIS_SCALA_VERSION coverage test && ./sbt ++$TRAVIS_SCALA_VERSION coverageAggregate
+    - scala: 2.11.7
+      script: travis_retry ./sbt ++$TRAVIS_SCALA_VERSION coverage test && ./sbt ++$TRAVIS_SCALA_VERSION coverageAggregate
 
 after_success: 
   - bash <(curl -s https://codecov.io/bash)
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,16 +29,18 @@ before_script:
   - ./bin/travisci
   - travis_retry ./sbt ++$TRAVIS_SCALA_VERSION update
 
-
 script: "echo no op"
 
 matrix:
   include:
     - scala: 2.10.6
-      # the 'scripted' sbt plugin tests are only run for 2.10
-      script: travis_retry ./sbt ++$TRAVIS_SCALA_VERSION scrooge-sbt-plugin/scripted && ./sbt ++$TRAVIS_SCALA_VERSION coverage test && ./sbt ++$TRAVIS_SCALA_VERSION coverageAggregate
-    - scala: 2.11.7
-      script: travis_retry ./sbt ++$TRAVIS_SCALA_VERSION coverage test && ./sbt ++$TRAVIS_SCALA_VERSION coverageAggregate
+      script: 
+        - travis_retry ./sbt ++$TRAVIS_SCALA_VERSION clean scrooge-sbt-plugin/scripted 
+    - scala: 2.11.8
+      script: 
+        - travis_retry ./sbt ++$TRAVIS_SCALA_VERSION clean coverage test coverageReport 
+        - travis_retry ./sbt ++$TRAVIS_SCALA_VERSION coverageAggregate
 
 after_success: 
   - bash <(curl -s https://codecov.io/bash)
+
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -285,7 +285,7 @@ object Scrooge extends Build {
       bintrayPublishSettings ++
       buildInfoSettings ++
       scriptedSettings
-  ).settings(
+    ).settings(
       sourceGenerators in Compile <+= buildInfo,
       buildInfoKeys := Seq[BuildInfoKey](name, version, scalaVersion, sbtVersion),
       buildInfoPackage := "com.twitter",
@@ -301,7 +301,7 @@ object Scrooge extends Build {
         "-Dlibthrift.version=" + libthriftVersion,
         "-Dfinagle.version=" + finagleVersion
       )
-  ).dependsOn(scroogeGenerator)
+    ).dependsOn(scroogeGenerator)
 
   lazy val scroogeLinter = Project(
     id = "scrooge-linter",
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -9,6 +9,7 @@ import sbtassembly.Plugin._
 import AssemblyKeys._
 import sbtbuildinfo.Plugin._
 import scoverage.ScoverageSbtPlugin
+import ScriptedPlugin._
 
 object Scrooge extends Build {
   val branch = Process("git" :: "rev-parse" :: "--abbrev-ref" :: "HEAD" :: Nil).!!.trim
@@ -277,7 +278,8 @@ object Scrooge extends Build {
     settings = Defaults.coreDefaultSettings ++
       settingsWithTwoTen ++
       bintrayPublishSettings ++
-      buildInfoSettings
+      buildInfoSettings ++
+      scriptedSettings
   ).settings(
       sourceGenerators in Compile <+= buildInfo,
       buildInfoKeys := Seq[BuildInfoKey](name, version, scalaVersion, sbtVersion),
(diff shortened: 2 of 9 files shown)


kevinoliver:
i'm pretty weak in terms of sbt knowledge so apologies in advance for my questions on the diff.

project/Build.scala
```diff
@@ -278,7 +280,9 @@ object Scrooge extends Build {
       publishMavenStyle := false,
       repository in bintray := "sbt-plugins",
       licenses += (("Apache-2.0", url("http://www.apache.org/licenses/LICENSE-2.0.html"))),
-      bintrayOrganization in bintray := Some("twittercsl")
+      bintrayOrganization in bintray := Some("twittercsl"),
+      scriptedBufferLog := false,
```
kevinoliver: can you leave comments documenting what these options mean

project/plugins.sbt
```diff
@@ -14,3 +14,6 @@ addSbtPlugin("net.virtual-void" % "sbt-dependency-graph" % "0.8.2")
 addSbtPlugin("org.scoverage" % "sbt-coveralls" % "1.0.0")
 addSbtPlugin("org.scoverage" % "sbt-scoverage" % "1.2.0")
 addSbtPlugin("pl.project13.scala" % "sbt-jmh" % "0.2.2")
+
+libraryDependencies += "org.scala-sbt" % "scripted-plugin" % sbtVersion.value
```
kevinoliver: am i correct that this is separate from the fix and enables testing of plugins? 

if so, can we split this into to two separate PRs?

scrooge-sbt-plugin/src/sbt-test/scrooge/basic/build.sbt
```diff
@@ -0,0 +1,10 @@
+
+scalaVersion := "2.11.8"
```
kevinoliver: i thought sbt plugins were 2.10 only?

scrooge-sbt-plugin/src/sbt-test/scrooge/basic/build.sbt
```diff
@@ -0,0 +1,10 @@
...
+
+enablePlugins(BuildInfoPlugin)
+
+libraryDependencies ++= Seq(
```
kevinoliver: these versions not correct when we are doing development. can we share the code thats in `Build.scala` for setting these?

scrooge-sbt-plugin/src/sbt-test/scrooge/basic/test
```diff
@@ -0,0 +1,3 @@
+> compile
+$ exists target/scala-2.11/src_managed/main/thrift/com/twitter/demo/User.scala
+$ exists target/scala-2.11/src_managed/main/sbt-buildinfo/BuildInfo.scala
```
kevinoliver: this is pretty cool.

scrooge-sbt-plugin/src/sbt-test/scrooge/basic/build.sbt
joprice: This version is the version of a test project using this plugin.

project/plugins.sbt
joprice: Yes this enables testing of the plugin, but the test I added is for this specific fix. You want me to separate that into another pr with a project that doesn't test this issue first, just a simple test that the plugin generates a scala file?

scrooge-sbt-plugin/src/sbt-test/scrooge/basic/build.sbt
joprice: Yea I was hoping for some direction on what to do with these. I thought it might be ok if the plugin is backwards compatible, but if that is not the case, we could export some more props for these versions.d

kevinoliver: yeah, i think we need it to work the same as it does in `Build.scala`.

scrooge-sbt-plugin/src/sbt-test/scrooge/basic/build.sbt
kevinoliver: got it, thanks.

project/plugins.sbt
kevinoliver: i see. ok lets leave them bundled.

project/Build.scala
joprice: The documentation for scripted is here: http://www.scala-sbt.org/0.13/docs/Testing-sbt-plugins.html. Do you want me to just put that link in a comment?

joprice:
@kevinoliver I updated the scripted test to use `sys.props` for dependency versions.

project/Build.scala
kevinoliver: that'd be awesome. thanks.

kevinoliver:
mind adding that one doc link, and then this lgtm. thanks @joprice

joprice:
Added the doc link. One more thing to note is that you have to run `scrooge-sbt-plugin/scripted` to kick off the tests. Should that be added to `.travis.yml`? It seems like the travis builds have been failing for a while.

kevinoliver:
yeah please add it to `.travis.yml`, thanks. 

i /think/ travis is back to working now, we had some sbt dependency issues we were working through.

joprice:
It only makes sense to run `scripted` on travis for scala 2.10, since it's an sbt plugin. So I would add it to `script` in `.travis.yml` without `++$TRAVIS_SCALA_VERSION` and it will get run twice. I haven't used it, but it looks like we could make use of the 'matrix' configuration and have a different command for each version: https://github.com/twitter/summingbird/blob/master/.travis.yml#L12

mosesn:
I think doing the matrix style or the "run the same thing twice" are both OK.  As long as it doesn't get in our way, I'm happy.

.travis.yml
```diff
@@ -34,5 +34,14 @@ before_script:
   - ./bin/travisci
   - travis_retry ./sbt ++$TRAVIS_SCALA_VERSION update
 
-script: ./sbt ++$TRAVIS_SCALA_VERSION coverage test && ./sbt ++$TRAVIS_SCALA_VERSION coverageAggregate
+script: "echo no op"
```
cacoco: I think your diff is out of date. We've switched to use codecov.io in the latest code (instead of coveralls). You may want to re-base the branch.

joprice:
Thanks! I'll update.

mosesn:
@joprice this looks good to me!  I'm going to try to resolve the merge conflicts and merge this in.  Thanks for the PR!

joprice:
Thanks! Unfortunately, I wasn't able to get this to pass CI :-(

mosesn:
@joprice no worries, we've had a lot of CI problems, so I'll see what I can do on our end.

mosesn:
@joprice I might be running into the same thing you did when trying to get CI to work.  was it problems around finding a copy of sbt 0.13.9 for the test?

joprice:
I don't remember the exact error sorry.

mosesn:
@joprice OK, this unfortunately looks like it's going to be something of a doozy because of the way scripted works, and it doesn't look like the scripted maintainers are very [responsive][0].  I'm going on vacation for a week, and will try again when I come back, but we might have to rip the test out of it to merge it in until we can figure out how to get scripted to work in our environment.

[0]: https://github.com/sbt/sbt/issues/2685

mosesn:
@joprice I think getting scripted to work with our stuff is going to be more trouble than it's worth for now.  I think I'm going to rip out the testing and then merge it in, this has taken way too long.  Sorry, I'll work in the future on getting scripted working.

joprice:
No problem. I just enabled the new 'allow edits from maintainers'.

jcrossley:
Hey @joprice, wanted to give you an update on this. Unfortunately, we can't merge it in yet (even without the scripted plugin and tests), because it's causing issues with the [Finatra ](https://github.com/twitter/finatra) framework. Trying to sort this out and will update you with progress. Thanks for contributing!

scrooge-sbt-plugin/src/main/scala/com/twitter/ScroogeSBT.scala
```diff
@@ -171,7 +171,7 @@ object ScroogeSBT extends AutoPlugin {
     scroogeDefaultJavaNamespace := "thrift",
     scroogeThriftSourceFolder <<= (sourceDirectory) { _ / "thrift" },
     scroogeThriftExternalSourceFolder <<= (target) { _ / "thrift_external" },
-    scroogeThriftOutputFolder <<= (sourceManaged) { identity },
+    scroogeThriftOutputFolder in Compile := (sourceManaged in Compile).value / "thrift",
```
cacoco: As the comments above this section point out the configuration here should be in both Compile *and* test configuration. Hence this is a breaking change since it sets the output folder in the Compile configuration only. This probably needs to be modified to be:

```
scroogeThriftOutputFolder in Compile := (sourceManaged in Compile).value / "thrift",
scroogeThriftOutputFolder in Test := (sourceManaged in Test).value / "thrift",
```

cacoco:
@joprice we're going to apply the change internally and get this merged as soon as we can. The change is under review internally. Thanks!

cacoco:
@joprice  this has been merged internally and should show up upon the next sync to Github. Thanks!

cacoco:
Merged in e604a7212e5eb73445a47a36223ea9f26e419607

Issue #236, opened 2016-05-18
2.12-ification


mosesn:
In order to support 2.12, we need to add crossScalaVersions to Build.scala, and make sure everything compiles.  Not every scala library will be built against the current snapshot, so we may have to support different versions from 2.11 to 2.12.  If part of an API has churned (scalacheck in particular is susceptible to this) we may need to upgrade the 2.11 library version to make progress.  If that happens, contact me (@mosesn) and I'll navigate the upgrade for you–it's a little complicated because we will need to upgrade all of our internal libraries, not just scrooge.  Please note that most of this upgrade will not be feasible until we have a portion of util already upgraded.  We are not upgrading `scrooge-sbt-plugin`, because sbt is not migrating to [2.12 anytime soon](https://github.com/sbt/sbt/issues/2389).
- [x] scrooge-benchmark
- [x] scrooge-core
- [x] scrooge-generator
- [ ] scrooge-generator-tests
- [x] scrooge-linter
- [x] scrooge-maven-plugin
- [x] scrooge-serializer

If you want to claim a module, please let us know in this thread!  I'll annotate different modules once they're done.

Probably a good way to start is for someone to try add 2.12 support for all of them, and then report back which ones still fail, so we know which ones are tricky and which ones work trivially.

taylorleese:
Related: https://github.com/twitter/util/issues/162

mosesn:
We will need to make sure that mustache scala-extensions-2.11 works against scala 2.12.  We also need to upgrade to mustache 0.9.x (OK now that we're on jdk 8).

jeffreyolchovy:
I'll take care of this one as well. See https://github.com/twitter/scrooge/pull/237 for current status.

mosesn:
Almost everything here has been tackled by https://github.com/twitter/scrooge/pull/237.  scrooge-generator-tests is still waiting on finagle.  Thanks @jeffreyolchovy!

sritchie:
Hey @mosesn, what's the status here? @johnynek and I are churning through the Summingbird dependencies; util looks like it's upgraded, but we're still waiting on Scrooge and Finagle. Do you have an ETA for a 2.12 build here?

Thanks!

mosesn:
@sritchie scrooge is already on 2.12, so you shouldn't be blocked here.  the only thing that's not is the tests.  http://search.maven.org/#search%7Cga%7C1%7Cscrooge-core_2.12

[...]

PR #237, opened 2016-05-29
2.12-ification

jeffreyolchovy:
Cross build for Scala 2.12
    
Problem
    
There exists a need to publish artifacts that are binary compatibile with Scala 2.12.
    
Solution
    
- Modify sbt build defintion to cross build for 2.12.
- Modify main and test sources to compile against 2.12.
- Ensure the use of scoverage artifacts that exist when targeting 2.12.
- Update scalatest and scalacheck libraries, when necessary, to versions that support 2.12.
- Update scopt library to a version that supports 2.12.
    
Result
    
- When targeting 2.12, the scoverage artifacts will still use their 2.11 versions.
- By default, scoverage is disabled when targeting 2.12 and for the scrooge-sbt-plugin.
- 2.12 compatible versions of scalatest and scalacheck are used when targeting 2.12 only.

<hr/>

I currently tested this without the scrooge-generator-tests. So far, with this changeset, all successfully compiles against 2.11.x and 2.12.x. My next steps are to ensure the scrooge-generator-tests work correctly and that all tests succeed (so consider a WIP).


project/Build.scala
```diff
@@ -18,7 +18,7 @@ object Scrooge extends Build {
 
   // To build the develop branch you need to publish util, and finagle locally:
   // 'git checkout develop; sbt publishLocal' to publish SNAPSHOT versions of these projects.
-  val utilVersion = "6.34.0" + suffix
+  val utilVersion = "6.35.0" + suffix
```
mosesn: not sure we should be changing this?

project/Build.scala
```diff
@@ -57,13 +57,29 @@ object Scrooge extends Build {
...
 
     libraryDependencies ++= Seq(
-      "org.scalatest" %% "scalatest" % "2.2.4" % "test",
-      "org.scalacheck" %% "scalacheck" % "1.12.2" % "test",
+      scalatestLib(scalaVersion.value) % "test",
+      scalacheckLib(scalaVersion.value) % "test",
       "junit" % "junit" % "4.12" % "test"
     ),
+
+    libraryDependencies := {
```
mosesn: instead of this, can we explicitly request scoverage 2.11 in the addSbtPlugin section?  Here's an example of what might be a simpler way to do this:

https://github.com/sbt/sbt/blob/83f35b18d348d36611570642833976d0f6520a60/sbt/src/sbt-test/dependency-management/override2/build.sbt#L3

project/Build.scala
```diff
@@ -205,8 +238,9 @@ object Scrooge extends Build {
     name := "scrooge-generator",
     libraryDependencies ++= Seq(
       "org.apache.thrift" % "libthrift" % libthriftVersion,
-      "com.github.scopt" %% "scopt" % "3.3.0",
-      "com.github.spullara.mustache.java" % "compiler" % "0.8.18",
+      "com.github.scopt" %% "scopt" % "3.4.0",
+      "com.google.guava" % "guava" % "19.0",
```
mosesn: we're still on 16.0.1, unfortunately

project/Build.scala
```diff
@@ -205,8 +238,9 @@ object Scrooge extends Build {
     name := "scrooge-generator",
     libraryDependencies ++= Seq(
       "org.apache.thrift" % "libthrift" % libthriftVersion,
-      "com.github.scopt" %% "scopt" % "3.3.0",
-      "com.github.spullara.mustache.java" % "compiler" % "0.8.18",
```
mosesn: I'm not sure we can move off of this yet.  is there a reason we have to?

mosesn:
Thanks!  I'm pretty excited by how fast all of this moving.

project/Build.scala
jeffreyolchovy: not sure how you handle this internally, my thoughts were that since it depends on the 2.12 branch that has not been released it would be: increment latest release version and add suffix. if i don't increment i am essentially publishing a local snapshot version that already has a corresponding release (6.34.0).

project/Build.scala
jeffreyolchovy: nope, no direct reason, though it is possible if that's the path you want to take. i was just addressing your comment here: https://github.com/twitter/scrooge/issues/236#issuecomment-220889358

project/Build.scala
jeffreyolchovy: nice find! going to try this out. if it works we can apply the same strategy to util and the rest that need the upgrade.

jeffreyolchovy: Unfortunately, this didn't work. That syntax only changes the resolution path for the plugin itself. For example using the following in `plugins.sbt`:
```
addSbtPlugin("org.scoverage" % "sbt-scoverage" % "1.3.5", sbtVersion = "0.13", scalaVersion = "2.11")
```
would result in the following error:
```
[info] Resolving org.scoverage#sbt-scoverage;1.3.5 ...
[warn] 	module not found: org.scoverage#sbt-scoverage;1.3.5
[warn] ==== typesafe-ivy-releases: tried
[warn]   https://repo.typesafe.com/typesafe/ivy-releases/org.scoverage/sbt-scoverage/scala_2.11/sbt_0.13/1.3.5/ivys/ivy.xml
[warn] ==== sbt-plugin-releases: tried
[warn]   https://repo.scala-sbt.org/scalasbt/sbt-plugin-releases/org.scoverage/sbt-scoverage/scala_2.11/sbt_0.13/1.3.5/ivys/ivy.xml
...
```
The only configuration that works would be:
```
addSbtPlugin("org.scoverage" % "sbt-scoverage" % "1.3.5", sbtVersion = "0.13", scalaVersion = "2.10")
```
which gives us the same result as `addSbtPlugin` without the additional parameters.

However, the problem of the non-existent-for-the-current-version-of-scala, automatically injected library dependencies that are included in the modules as a result of the plugin remains. Right now, mapping over the libraryDependencies as done both here and in `util` are the best way that I found to resolve this so far.

project/Build.scala
mosesn: we change it as part of our release process, so we should be OK.

project/Build.scala
mosesn: ah, sorry, to be clear I meant move internally.  we need to do it ourselves because we have to maintain two build systems, the internal monorepo, and the public sbt.  sorry!

mosesn:
@jeffreyolchovy looks like we can't upgrade scoverage yet–it seems like it breaks builds:

https://github.com/twitter/util/pull/166

taylorleese:
@mosesn @cacoco Can we just run scoverage on 2.11 and omit it for 2.12?

jeffreyolchovy:
@taylorleese The conditional enabling of scoverage is already included in both this changeset and the `util` changeset that is (still?) experiencing problems -- see https://github.com/jeffreyolchovy/scrooge/blob/ac15b5ffe585c712c653e75d378048ea8ca7cd92/project/Build.scala#L79-L81

project/Build.scala
```diff
@@ -205,7 +238,7 @@ object Scrooge extends Build {
     name := "scrooge-generator",
     libraryDependencies ++= Seq(
       "org.apache.thrift" % "libthrift" % libthriftVersion,
-      "com.github.scopt" %% "scopt" % "3.3.0",
+      "com.github.scopt" %% "scopt" % "3.4.0",
```
mosesn: I'm bumping this internally too.

project/plugins.sbt
```diff
@@ -12,5 +12,5 @@ addSbtPlugin("com.typesafe.sbt" % "sbt-site" % "0.8.1")
 addSbtPlugin("me.lessis" % "bintray-sbt" % "0.1.1")
 addSbtPlugin("net.virtual-void" % "sbt-dependency-graph" % "0.8.2")
 addSbtPlugin("org.scoverage" % "sbt-coveralls" % "1.0.0")
-addSbtPlugin("org.scoverage" % "sbt-scoverage" % "1.2.0")
+addSbtPlugin("org.scoverage" % "sbt-scoverage" % "1.3.5")
```
mosesn: looks like we probably shouldn't bump scoverage for now.

project/Build.scala
```diff
@@ -162,6 +179,22 @@ object Scrooge extends Build {
     )
   )
 
+  def scalacheckLib(scalaVersion: String) = {
```
mosesn: I think we bumped this and scalatest so that we don't need to do this anymore.

jeffreyolchovy:
will rebase against upstream and update this in response to your comments shortly

jeffreyolchovy:
@mosesn rebased and updated.. still have a dependency issue since the scrooge-generator-tests depend on `finagle-thrift` so this comment in the PR description still applies:

> I currently tested this without the scrooge-generator-tests. So far, with this changeset, all successfully compiles against 2.11.x and 2.12.x. My next steps are to ensure the scrooge-generator-tests work correctly and that all tests succeed (so consider a WIP).

let me know if you have a good way of resolving these types of dependencies. and obviously, this also requires me locally publishing (+publishLocal) the HEAD of `util`.

mosesn:
Let's ignore the tests for now.  We can fix them up in a subsequent commit, and we don't need to publish with them.

mosesn:
Sweet, LGTM!

mosesn:
Actually, can you kick out the scrooge-generator-tests before we ship it?  We want to make sure the build stays green.

jeffreyolchovy:
not sure if necessary, all green as of now with scrooge-generator-tests in there.

mosesn:
OK, cool!  Works for me.

project/Build.scala
```diff
@@ -112,6 +129,7 @@ object Scrooge extends Build {
     sharedSettingsWithoutScalaVersion ++
     Seq(
       scalaVersion := "2.11.8",
+      crossScalaVersions := scalaVersion.value +: Seq("2.12.0-M4"),
```
cacoco: Can we keep this consistent with how we do this in other projects? I.e., how we do it in util [here](https://github.com/twitter/util/blob/develop/project/Build.scala#L30).

jeffreyolchovy:
^ sensitive test? it waffled from success to failure from the trivial crossScalaVersions change..

mosesn:
Hmm, it looks like it failed in retries too . . . that's unusual.  Can you try it locally and see how it works?  I don't know how to rebuild a travisci build.

mosesn:
OK, I figured it out, although it looks like it clobbered the old logs =(

mosesn:
@jeffreyolchovy hey, this was merged internally here: https://github.com/twitter/scrooge/commit/12472dbf784d867299af964672a177c5c6cc5c4d!  Thanks so much =)

I'm going to try to merge in the ostrich one too, so then we should be able to tackle finagle!

Issue #239, opened 2016-07-26
sbt-scrooge-plugin doesn't include generated sources in packageSrc


chessman:
sbt-scrooge-plugin doesn't include generated sources in packageSrc.

### Expected behavior

sbt-scrooge-plugin includes generated sources from `scroogeThriftOutputFolder` in sources jar.

### Actual behavior

sbt-scrooge-plugin generates empty sources jar.

### Steps to reproduce the behavior

build.sbt
``` scala
import sbt.Keys.{artifactPath, libraryDependencies, mainClass, managedClasspath, name, organization, packageBin, resolvers, version}

lazy val projectName = "scrooge-bug"
lazy val projectVersion = "0.0.1"

organization := "com.example"

scalaVersion := "2.11.8"

version := s"$projectVersion-SNAPSHOT"

name := s"$projectName"

libraryDependencies ++= Seq(
  "org.apache.thrift" % "libthrift" % "0.9.3",
  "com.twitter" %% "scrooge-core" % "4.5.0",
  "com.twitter" %% "scrooge-linter" % "4.5.0",
  "com.twitter" %% "finagle-thrift" % "6.33.0"
)
```

project/plugins.sbt
``` scala
addSbtPlugin("com.twitter" %% "scrooge-sbt-plugin" % "4.5.0")
```

src/main/thrift/test.thrift
``` thrift
namespace js test
#@namespace scala com.example.thrift.test

struct TestStruct {
    1: required i64 abc
    2: required string cba
}
```

1. sbt compile
2. sbt packageSrc
3. target/scala-2.11/scrooge-bug_2.11-0.0.1-SNAPSHOT-sources.jar doesn't include generated sources (target/scala-2.11/src_managed/main/com/example/thrift/test/TestStruct.scala).

mosesn:
@chessman hmm, this is really interesting.  thanks for the bug report!  Is this something we can change in the scrooge sbt plugin, or do you have to do it on your side?

It looks like there's an explanation [here][0].

so it looks like we would do something like:

```scala
mappings in (Compile, packageSrc) <+= baseDirectory.map { base =>
   (base / "target" / "scala-2.11" / "src_managed" ) -> "src_managed"
}
```

but would it work if we added it to [projectSettings][1]?

If it turns out we can fix it in scrooge, would you be interested in making a PR?

[0]: http://www.scala-sbt.org/0.12.1/docs/Howto/package.html#contents
[...]

PR #240, opened 2016-07-27
scrooge-sbt-plugin: Include generated sources in packageSrc

chessman:
Issue: https://github.com/twitter/scrooge/issues/239.

Problem

sbt-scrooge-plugin generates empty sources jar.

Solution

Include all generated sources in packageSrc.

Result

Now it is possible to browse generated sources with modern IDE.

### DIFF
--- a/scrooge-sbt-plugin/src/main/scala/com/twitter/ScroogeSBT.scala
+++ b/scrooge-sbt-plugin/src/main/scala/com/twitter/ScroogeSBT.scala
@@ -285,10 +285,15 @@ object ScroogeSBT extends AutoPlugin {
     }
   }
 
+  private[this] val generatedSources = mappings in (Compile, packageSrc) ++= {
+    val thriftOutputFolder = (scroogeThriftOutputFolder in Compile).value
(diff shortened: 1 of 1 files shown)


scrooge-sbt-plugin/src/main/scala/com/twitter/ScroogeSBT.scala
```diff
@@ -285,10 +285,15 @@ object ScroogeSBT extends AutoPlugin {
     }
   }
 
+  val generatedSources = mappings in (Compile, packageSrc) ++= {
```
mosesn: `private[this]` ?

chessman:
I've changed it to private[this] but most of the above vars in the object are not private =).

I checked it on my scala projects. It creates jar with all generated files and works correctly with ENSIME.

mosesn:
excellent!  we'll work on getting it merged internally, thanks!

mosesn:
@chessman this was merged internally, it should show up on the `develop` branch Monday.  Thanks for the contribution!

mosesn:
@chessman this made it to `develop` here: https://github.com/twitter/scrooge/commit/572609efc41575590a8790b13241b374f26e9c9e thanks for your contribution!

PR #241, opened 2016-07-28
Fix TReusableMemoryTransport reallocation on every reset

lightyang:
Problem

TReusableMemoryTransport is used in generated code to reduce byte[] allocation.

``` scala
  private[this] val tlReusableBuffer = new ThreadLocal[TReusableMemoryTransport] {
    override def initialValue() = TReusableMemoryTransport(512)
  }

  private[this] def reusableBuffer: TReusableMemoryTransport = {
    val buf = tlReusableBuffer.get()
    buf.reset()
    buf
  }

  private[this] val resetCounter = stats.scope("buffer").counter("resetCount")

  private[this] def resetBuffer(trans: TReusableMemoryTransport): Unit = {
    if (trans.currentCapacity > maxThriftBufferSize) {
      resetCounter.incr()
      tlReusableBuffer.remove()
    }
  }
```

However, due to [the way TByteArrayOutputStream is implemented](https://github.com/apache/thrift/blob/master/lib/java/src/org/apache/thrift/TByteArrayOutputStream.java#L46), `buf.reset()` will allocate a new underlying byte[] if current capacity exceeds the initial size (512).

The expected behavior is the underlying byte[] is only reallocated when current capacity exceeds `maxThriftBufferSize`, which is what the above `resetBuffer` method does.

Solution

Implement a `TUnboundedByteArrayOutputStream` class that doesn't do any reallocation and leave that to the `resetBuffer` method.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -5,6 +5,9 @@ as it is included in Scrooge's user's guide.
 4.x
 -----
 
+- scrooge-core: `c.t.scrooge.TReusableMemoryTransport` now uses TUnboundedByteArrayOutputStream
+  instead of TByteArrayOutputStream to avoid buffer reallocation on reset.
+
 4.8.0
 ------
 
@@ -19,7 +22,6 @@ Breaking API Changes
 * Builds are now only for Java 8 and Scala 2.11. See the
   `blog post <https://finagle.github.io/blog/2016/04/20/scala-210-and-java7/>`_
   for details. ``RB_ID=828898``
-
 
 ~~~~~
 
--- a/scrooge-benchmark/src/main/scala/com/twitter/scrooge/benchmark/TByteArrayOutputStream.scala
+++ b/scrooge-benchmark/src/main/scala/com/twitter/scrooge/benchmark/TByteArrayOutputStream.scala
@@ -1,0 +1,59 @@
+package com.twitter.scrooge.benchmark
+
+import com.twitter.scrooge.TUnboundedByteArrayOutputStream
+import java.util.concurrent.TimeUnit
+import org.apache.thrift.TByteArrayOutputStream
+import org.openjdk.jmh.annotations._
+
+object TByteArrayOutputStreamBenchmark {
+  @State(Scope.Thread)
+  class InputState {
+    val bytes: Int = 16
+    val buf = new Array[Byte](bytes)
+  }
+
+  @State(Scope.Benchmark)
+  class TByteArrayOutputStreamState {
+    @Param(Array("1024", "10240", "51200"))
+    var initialSize: Int = 1024
+
+    var stream: TByteArrayOutputStream = _
+
+    @Setup(Level.Trial)
+    def setup(): Unit = {
+      stream = new TByteArrayOutputStream(initialSize)
+    }
+  }
+
+  @State(Scope.Benchmark)
+  class TUnboundedByteArrayOutputStreamState {
+    @Param(Array("1024", "10240", "51200"))
+    var initialSize: Int = 1024
+
+    var stream: TByteArrayOutputStream = _
+
+    @Setup(Level.Trial)
+    def setup(): Unit = {
+      stream = new TUnboundedByteArrayOutputStream(initialSize)
+    }
+  }
+}
+
+@OutputTimeUnit(TimeUnit.MILLISECONDS)
+@BenchmarkMode(Array(Mode.Throughput))
+class TByteArrayOutputStreamBenchmark {
+  import TByteArrayOutputStreamBenchmark._
+
+  def run(in: Array[Byte], out: TByteArrayOutputStream): Unit = {
+    out.write(in, 0, in.size)
+    out.reset()
+  }
+
+  @Benchmark
+  def timeTByteArrayOutputStream(in: InputState, out: TByteArrayOutputStreamState) =
+    run(in.buf, out.stream)
+
+  @Benchmark
+  def timeTUnboundedByteArrayOutputStream(in: InputState, out: TUnboundedByteArrayOutputStreamState) =
+    run(in.buf, out.stream)
+}
--- a/scrooge-core/src/main/scala/com/twitter/scrooge/TReusableMemoryTransport.scala
+++ b/scrooge-core/src/main/scala/com/twitter/scrooge/TReusableMemoryTransport.scala
@@ -6,7 +6,7 @@ import org.apache.thrift.TByteArrayOutputStream
 object TReusableMemoryTransport {
 
   def apply(initialSize: Int = 512): TReusableMemoryTransport = {
-    new TReusableMemoryTransport(new TByteArrayOutputStream(initialSize))
+    new TReusableMemoryTransport(new TUnboundedByteArrayOutputStream(initialSize))
   }
 
 }
--- a/scrooge-core/src/main/scala/com/twitter/scrooge/TUnboundedByteArrayOutputStream.scala
+++ b/scrooge-core/src/main/scala/com/twitter/scrooge/TUnboundedByteArrayOutputStream.scala
@@ -1,0 +1,14 @@
+package com.twitter.scrooge
+
+import org.apache.thrift.TByteArrayOutputStream
+
+/**
+ * A version of ByteArrayOutputStream that exposes the internal buffer.
+ * It's used to replace TByteArrayOutputStream so that the internal buffer won't be reallocated
+ * on reset.
+ */
+private[scrooge] class TUnboundedByteArrayOutputStream(size: Int) extends TByteArrayOutputStream(size) {
+  override def reset(): Unit = {
+    count = 0
+  }
+}
--- a/scrooge-generator-tests/src/test/scala/com/twitter/scrooge/TReusableMemoryTransportSpec.scala
+++ b/scrooge-generator-tests/src/test/scala/com/twitter/scrooge/TReusableMemoryTransportSpec.scala
@@ -30,4 +30,18 @@ class TReusableMemoryTransportSpec extends Spec {
     trans.numWrittenBytes must be(5)
   }
 
+  "does not reallocate buffer on reset()" in {
+    val cap = 5
+    val trans = TReusableMemoryTransport(cap)
+
+    val stringInBytes = "abcdefghij".getBytes("UTF-8")
+    trans.write(stringInBytes)
+    trans.numWrittenBytes must be(10)
+
+    val oldBuf = trans.getArray
+    trans.reset()
+    trans.numWrittenBytes must be(0)
+    trans.getArray must be theSameInstanceAs oldBuf
+  }
+
 }


scrooge-core/src/main/scala/com/twitter/scrooge/TUnboundeByteArrayOutputStream.scala
```diff
@@ -0,0 +1,9 @@
+package com.twitter.scrooge
+
+import java.io.ByteArrayOutputStream
+
+class TUnboundedByteArrayOutputStream(size: Int) extends ByteArrayOutputStream(size) {
```
mosesn: could you add a comment as to why we need this?

scrooge-core/src/main/scala/com/twitter/scrooge/TReusableMemoryTransport.scala
```diff
@@ -16,7 +15,7 @@ object TReusableMemoryTransport {
  * object allocations.
  */
 class TReusableMemoryTransport(
-    baos: TByteArrayOutputStream)
+    baos: TUnboundedByteArrayOutputStream)
```
mosesn: this is a breaking API change, can you add a note in the scrooge changelog?

mosesn:
This looks good overall.  Do you think there's a way that we could change this so that if folks wanted the old behavior, they could still have it?

lightyang:
@mosesn Thanks for the review. I've updated the diff.
To get the old behavior, we can set maxThriftBufferSize to 512 when using the generated service, or maxReusableBufferSize when using ThriftStructSerializer. I think the new behavior should become the default as the old behavior struck me as totally unexpected and very inefficient.

kevinoliver:
Thanks for the PR. Looks like the implementation of `TByteArrayOutputStream` changed since libthrift 0.5 which is what Scrooge is using (http://grepcode.com/file/repo1.maven.org/maven2/org.apache.cassandra.deps/libthrift/0.5.0/org/apache/thrift/TByteArrayOutputStream.java).

That said, I realize that some people use newer versions of libthrift so we should still find a way forward for you.

Couple of thoughts on how to get there:

1. Can we keep the `TReusableMemoryTransport` API taking a `TByteArrayOutputStream` and have your `TUnboundedByteArrayOutputStream` extend `TByteArrayOutputStream` and then use that in the factory method (`TReusableMemoryTransport.apply`)?
1. Reduce the visibility of `TUnboundedByteArrayOutputStream` to `private[scrooge]` to minimize API surface area.
1. Please add a JMH benchmark for this change in `scrooge-benchmark` to show the performance and allocation rate improvements.

mosesn:
@lightyang do you think you'll get a chance to take another look at this any time soon?

lightyang:
@mosesn ah, sorry, kinda forget about this. I will try to address those issues in this week or next

lightyang:
@mosesn @kevinoliver I made `TUnboundedByteArrayOutputStream ` a subclass of `TByteArrayOutputStream` and preserved the API. Let me check out how to write a benchmark...

lightyang:
added jmh benchmark

scrooge-benchmark/src/main/scala/com/twitter/scrooge/benchmark/TByteArrayOutputStream.scala
```diff
@@ -0,0 +1,41 @@
...
+}
+
+@OutputTimeUnit(TimeUnit.MILLISECONDS)
+@BenchmarkMode(Array(Mode.Throughput))
+class TByteArrayOutputStreamBenchmark {
+  import TByteArrayOutputStreamBenchmark._
+
+  @Benchmark
+  def timeTByteArrayOutputStream(state: TByteArrayOutputStreamState) =
+    state.run(new TByteArrayOutputStream(state.initialSize))
```
kevinoliver: JMH has a very opinionated style of writing code and this goes against that.

instead, what you can do:
* lift the output streams into your `@State` class. then "inject" a state instance into your `Benchmarks`.  as an example, look at https://github.com/twitter/finagle/blob/develop/finagle-benchmark/src/main/scala/com/twitter/finagle/stats/StatsReceiverBenchmark.scala
* lift the creation of the `Array[Byte]` into the `State` as you want to remove that work from the test.
* You do not want a loop in the `run` method. it should only call `write` and `reset`. JMH will "do the looping".

CHANGES
```diff
@@ -20,6 +20,8 @@ Breaking API Changes
   `blog post <https://finagle.github.io/blog/2016/04/20/scala-210-and-java7/>`_
   for details. ``RB_ID=828898``
 
+* scrooge-core: `c.t.scrooge.TReusableMemoryTransport` now uses TUnboundedByteArrayOutputStream
```
kevinoliver: please move this up to 4.x since 4.8 already shipped since you started this PR.

also, i think this is a "New Feature" instead of a "Breaking API Change", right?

scrooge-core/src/main/scala/com/twitter/scrooge/TUnboundedByteArrayOutputStream.scala
```diff
@@ -0,0 +1,14 @@
...
+
+import org.apache.thrift.TByteArrayOutputStream
+
+/**
+ * A version of ByteArrayOutputStream that exposes the internal buffer.
+ * It's used to replace TByteArrayOutputStream so that the internal buffer won't be reallocated
+ * on reset.
+ */
+private[scrooge] class TUnboundedByteArrayOutputStream(size: Int) extends TByteArrayOutputStream(size) {
+  override def reset() {
```
kevinoliver: style nitpick: scala is moving away from procedure style methods. can you annotate the return type? `: Unit = {`

lightyang: sure

lightyang:
@kevinoliver I ran jmh with a small heap (`jmh:run -i 50 -wi 10 -f1 -t4 -prof gc --jvmArgs "-Xmx32M"`) and got the following results. Though I then realized scrooge is using libthrift 0.5.0 and these two streams should behave more or less identically.
```
[info] Benchmark                                                                                    (initialSize)   Mode  Cnt      Score      Error   Units
[info] TByteArrayOutputStreamBenchmark.timeTByteArrayOutputStream                                            1024  thrpt   50  14354.305 ±  833.495  ops/ms
[info] TByteArrayOutputStreamBenchmark.timeTByteArrayOutputStream:·gc.alloc.rate                             1024  thrpt   50      0.008 ±    0.009  MB/sec
[info] TByteArrayOutputStreamBenchmark.timeTByteArrayOutputStream:·gc.alloc.rate.norm                        1024  thrpt   50      0.001 ±    0.001    B/op
[info] TByteArrayOutputStreamBenchmark.timeTByteArrayOutputStream:·gc.count                                  1024  thrpt   50        ≈ 0             counts
[info] TByteArrayOutputStreamBenchmark.timeTByteArrayOutputStream                                           10240  thrpt   50  15286.270 ± 1017.485  ops/ms
[info] TByteArrayOutputStreamBenchmark.timeTByteArrayOutputStream:·gc.alloc.rate                            10240  thrpt   50      0.013 ±    0.033  MB/sec
[info] TByteArrayOutputStreamBenchmark.timeTByteArrayOutputStream:·gc.alloc.rate.norm                       10240  thrpt   50      0.001 ±    0.002    B/op
[info] TByteArrayOutputStreamBenchmark.timeTByteArrayOutputStream:·gc.churn.PS_Eden_Space                   10240  thrpt   50      0.169 ±    0.593  MB/sec
[info] TByteArrayOutputStreamBenchmark.timeTByteArrayOutputStream:·gc.churn.PS_Eden_Space.norm              10240  thrpt   50      0.009 ±    0.031    B/op
[info] TByteArrayOutputStreamBenchmark.timeTByteArrayOutputStream:·gc.churn.PS_Survivor_Space               10240  thrpt   50     ≈ 10⁻⁴             MB/sec
[info] TByteArrayOutputStreamBenchmark.timeTByteArrayOutputStream:·gc.churn.PS_Survivor_Space.norm          10240  thrpt   50     ≈ 10⁻⁵               B/op
[info] TByteArrayOutputStreamBenchmark.timeTByteArrayOutputStream:·gc.count                                 10240  thrpt   50      1.000             counts
[info] TByteArrayOutputStreamBenchmark.timeTByteArrayOutputStream:·gc.time                                  10240  thrpt   50      1.000                 ms
[info] TByteArrayOutputStreamBenchmark.timeTByteArrayOutputStream                                           51200  thrpt   50  14863.655 ±  792.143  ops/ms
[info] TByteArrayOutputStreamBenchmark.timeTByteArrayOutputStream:·gc.alloc.rate                            51200  thrpt   50      0.008 ±    0.013  MB/sec
[info] TByteArrayOutputStreamBenchmark.timeTByteArrayOutputStream:·gc.alloc.rate.norm                       51200  thrpt   50      0.001 ±    0.001    B/op
[info] TByteArrayOutputStreamBenchmark.timeTByteArrayOutputStream:·gc.churn.PS_Eden_Space                   51200  thrpt   50      0.170 ±    0.594  MB/sec
[info] TByteArrayOutputStreamBenchmark.timeTByteArrayOutputStream:·gc.churn.PS_Eden_Space.norm              51200  thrpt   50      0.012 ±    0.043    B/op
[info] TByteArrayOutputStreamBenchmark.timeTByteArrayOutputStream:·gc.count                                 51200  thrpt   50      1.000             counts
[info] TByteArrayOutputStreamBenchmark.timeTByteArrayOutputStream:·gc.time                                  51200  thrpt   50      2.000                 ms
[info] TByteArrayOutputStreamBenchmark.timeTUnboundedByteArrayOutputStream                                   1024  thrpt   50  15930.703 ±  611.981  ops/ms
[info] TByteArrayOutputStreamBenchmark.timeTUnboundedByteArrayOutputStream:·gc.alloc.rate                    1024  thrpt   50      0.029 ±    0.070  MB/sec
[info] TByteArrayOutputStreamBenchmark.timeTUnboundedByteArrayOutputStream:·gc.alloc.rate.norm               1024  thrpt   50      0.002 ±    0.005    B/op
[info] TByteArrayOutputStreamBenchmark.timeTUnboundedByteArrayOutputStream:·gc.count                         1024  thrpt   50        ≈ 0             counts
[info] TByteArrayOutputStreamBenchmark.timeTUnboundedByteArrayOutputStream                                  10240  thrpt   50  15441.676 ±  628.408  ops/ms
[info] TByteArrayOutputStreamBenchmark.timeTUnboundedByteArrayOutputStream:·gc.alloc.rate                   10240  thrpt   50      0.007 ±    0.015  MB/sec
[info] TByteArrayOutputStreamBenchmark.timeTUnboundedByteArrayOutputStream:·gc.alloc.rate.norm              10240  thrpt   50     ≈ 10⁻³               B/op
[info] TByteArrayOutputStreamBenchmark.timeTUnboundedByteArrayOutputStream:·gc.count                        10240  thrpt   50        ≈ 0             counts
[info] TByteArrayOutputStreamBenchmark.timeTUnboundedByteArrayOutputStream                                  51200  thrpt   50  14505.722 ±  754.041  ops/ms
[info] TByteArrayOutputStreamBenchmark.timeTUnboundedByteArrayOutputStream:·gc.alloc.rate                   51200  thrpt   50      0.005 ±    0.004  MB/sec
[info] TByteArrayOutputStreamBenchmark.timeTUnboundedByteArrayOutputStream:·gc.alloc.rate.norm              51200  thrpt   50     ≈ 10⁻³               B/op
[info] TByteArrayOutputStreamBenchmark.timeTUnboundedByteArrayOutputStream:·gc.count                        51200  thrpt   50        ≈ 0             counts
```

kevinoliver:
looks great @lightyang! thanks.

kevinoliver:
alrighty, just merged this in and it'll show up here in Monday's push to github. Thanks again for the patch and your patience on getting it right.

kevinoliver:
This got merged a few weeks back in 00dbfd7109ed. Thanks again.

Issue #243, opened 2016-08-16
Self-references leads to error


nollbit:
Self-references leads to error

### Expected behavior

Self-referencing structs should be supported

### Actual behavior

Self-referencing structs leads to `TypeNotFoundException`

### Steps to reproduce the behavior

```
namespace java com.test.thrift

struct Hello {
	1: optional Hello hello;
}
```
This is supported by later Thrift compilers. I'm not sure this is a bug in Scrooge since afaik it has never promised to be fully Thrift compatible.

kevinoliver:
Yeah, this is not supported today. This is the same as issue https://github.com/twitter/scrooge/issues/160 which has more commentary.

Issue #160, opened 2014-11-23
Add support for Tree/Recursive structs


zfy0701:
which has already been included in thrift 0.9.2:
https://issues.apache.org/jira/browse/THRIFT-2421

pavanch007:
Is anyone working on this scrooge feature?

kevinoliver:
@pavanch007 nobody here at twitter is working on it as far as i know.

pavanch007:
Thanks.. good to know. I will take this on.. do you have any pointers to start with this task?

kevinoliver:
@pavanch007 nice! i haven't spent any time thinking about this one so i don't have much in the way of advice.

best i can do is tell you what i'd do. i'd start by adding a recursive struct and service using that struct to a test idl in scrooge-generator. then see what blows up when you run the tests. once the parser and generator is not blowing up, i'd update a test case to use the recursive struct and verify the behavior.

we can try to work through this with you as you go forward.

fkorotkov:
@pavanch007, have you made any progress? Just curious :-)

pavanch007:
@fkorotkov Sorry for not circling back. This got de-prioritized and I'm not actively working on this.

mosesn:
This issue is related to https://github.com/twitter/scrooge/issues/85

taylorleese:
I believe this was closed in error so reopening.

mosesn:
This was fixed in #244 🎆

ihorkhavkin:
@mosesn is it a mistake to resolve this issue? see "This change doesn't yet support recursive types which is needed to fully resolve #160." on #244

mosesn:
@ihorkhavkin maybe!  what's the distinction between self-referential types and recursive types?

zfy0701:
recursive:
```
struct A {
  1 B b;
}

struct B {
  1 A a;
}
```
self referential
struct A {
  1 A a;
}

mosesn:
@zfy0701 got it, thanks!

folex:
Please do!

mosesn:
This was fixed by @folone here: https://github.com/twitter/scrooge/commit/2a731bbc49cdf11372cd1d001b9446451dcd84fe

folone:
Nice, thanks for pushing this @mosesn!

PR #244, opened 2016-09-20
Resolve self referencing types.

pinsri:
Problem

Scrooge currently doesn't support self-referencing types. This is useful for a number of cases.

Solution

Change the type resolver's scope to include the current type. Add unit tests. Updated gold file and test.thrift to include a self-referencing type. Fixed a bug in GoldFileTest.scala's diff calculation code where if genStr is a prefix of expected, line 108 (getStr(i)) will generate OutOfBoundsException (since the diff is after the genStr string).

Result

Self-referencing types no longer result in TypeNotFoundException. This should resolve https://github.com/twitter/scrooge/issues/243 and partially resolve https://github.com/twitter/scrooge/issues/160. This change doesn't yet support recursive types which is needed to fully resolve https://github.com/twitter/scrooge/issues/160.

This should be a backward compatible change since all the thrift files that are resolved correctly now will continue to get resolved correctly.

One caveat though is that we don't enforce the self-reference type to be optional or collection type. We just depend on the generated language's compiler to enforce that (i.e. we expect that compiler to generate an error if a self-referencing type is specified as required).

### DIFF
--- a/scrooge-generator-tests/src/test/resources/gold_file_input/gold.thrift
+++ b/scrooge-generator-tests/src/test/resources/gold_file_input/gold.thrift
@@ -21,6 +21,8 @@ struct Request {
   1: list<string> aList,
   2: set<i32> aSet,
   3: map<i64, i64> aMap,
+  4: optional Request aRequest,
+  5: list<Request> subRequests
 }
 
 struct Response {
--- a/scrooge-generator-tests/src/test/resources/gold_file_output_android/com/twitter/scrooge/test/gold/thriftandroid/Request.java
+++ b/scrooge-generator-tests/src/test/resources/gold_file_output_android/com/twitter/scrooge/test/gold/thriftandroid/Request.java
@@ -31,17 +31,23 @@ public class Request implements TBase<Request, Request._Fields>, java.io.Seriali
   private static final TField A_LIST_FIELD_DESC = new TField("aList", TType.LIST, (short)1);
   private static final TField A_SET_FIELD_DESC = new TField("aSet", TType.SET, (short)2);
   private static final TField A_MAP_FIELD_DESC = new TField("aMap", TType.MAP, (short)3);
+  private static final TField A_REQUEST_FIELD_DESC = new TField("aRequest", TType.STRUCT, (short)4);
+  private static final TField SUB_REQUESTS_FIELD_DESC = new TField("subRequests", TType.LIST, (short)5);
 
 
   private List<String> aList;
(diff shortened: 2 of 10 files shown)


kevinoliver:
@pinsri thanks! the change seems small enough. regardless, i'm nervous! 

to alleviate my paranoia, can you add 2 more tests? first where you add a self-reference to one of the IDLs under `scrooge-generator-tests/src/test/thrift`? second, modify a gold file IDL in `scrooge-generator-tests/src/test/resources/gold_file_input` so we can see what this does to the generated code.

thanks again.

pinsri:
let me try. thanks.

mosesn:
LGTM

pinsri:
Added 2 more tests. Adding to gold file input took way way more time than I expected. I eventually hacked up GoldFileTest.scala to print the full path of the generated file and not delete the temp directory and copied those files back (with some cleanup needed as well). Is there some simpler way to generate gold outputs that I'm missing?

kevinoliver:
Thanks @pinsri. Yeah, the gold file test is a pain. When I've done it in the past, I thought it logged the new output to stdout (see https://github.com/twitter/scrooge/blob/develop/scrooge-generator-tests/src/test/scala/com/twitter/scrooge/goldfile/GoldFileTest.scala#L115-L128). Maybe that wasn't working?

pinsri:
It was outputting to stdout, but copy pasting introduced unnecessary spaces
for me.

pinsri:
@kevinoliver : does the PR look ok to you now?

kevinoliver:
@pinsri yep! working on getting it merged in right now. thanks for the patch.

pinsri:
Great, thanks for your review, @kevinoliver

pinsri:
Also, do you know when the next release is going to be?

kevinoliver:
We are targeting the beginning of October for the next release.

kevinoliver:
Got this merged and it should show up on the `develop` branch this Monday. Thanks again.

mosesn:
@pinsri this made it to develop here: https://github.com/twitter/scrooge/commit/2271a73bda8d6633359f71b72c6d24b8fa61b830 thanks for the contribution!

pinsri:
@mosesn @kevinoliver Is the release going to be this week? Thanks.

kevinoliver:
@pinsri Yep, looking very likely. There is one OSS issue we are trying to get sorted out before hand.

PR #248, opened 2016-11-13
Struct unapply returns Option[TupleN]

jstultz:
Problem

Currently, the generated code for thrift structs defines an unapply
method that returns `Option[ProductN]`. This is inconsistent with
Scala's built in behavior when unapply methods are generated for case
classes, which return `Option[TupleN]`. A number of libraries work well
with the latter but not the former (e.g. scala.math.Ordering implicits,
Slick table mappings).

Solution

For consistency with the behavior of Scala case classes, and to improve
library interoperability, this modifies the generated code to return a
tuple rather than product instance. For consistency, the definition of
the ProductN types is also modified to use the `_root_` prefix.

Result

Library interoperability and consistency should be improved; impact on
existing code should be minimal as each TupleN class implements the
corresponding ProductN trait.

I was not sure exactly how best to handle the case where `tupleN` is called for a struct with arity 0 or greater than 22; for now I've just done the same as `productN`; it should never be used, but still doesn't quite feel like the best solution.

### DIFF
--- a/scrooge-generator-tests/src/test/resources/gold_file_output_scala/com/twitter/scrooge/test/gold/thriftscala/CollectionId.scala
+++ b/scrooge-generator-tests/src/test/resources/gold_file_output_scala/com/twitter/scrooge/test/gold/thriftscala/CollectionId.scala
@@ -279,7 +279,7 @@ object CollectionId extends ThriftStructCodec3[CollectionId] {
 
 trait CollectionId
   extends ThriftStruct
-  with scala.Product1[Long]
+  with _root_.scala.Product1[Long]
   with HasThriftStructCodec3[CollectionId]
   with java.io.Serializable
 {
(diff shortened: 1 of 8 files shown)


kevinoliver:
@jstultz this patch looks great! thanks.

kevinoliver:
Sorry for the delay on this. Working on it today.

kevinoliver:
ok unfortunately this is causing 5 other twitter tests to fail when i pull it in. they seem to be using the structs in match statements. along the lines of:

```
someMethodCallReturningThriftStruct match {
  case TheStructHere(arg1, ...) => // do something with it
}
```
i'll try to keep looking at this.

kevinoliver:
ok it was 2 issues:
# using mocks for thrift structs that were unnecessary
# an errant extra unapply on a thrift struct that was unnecessary

hoping to get this in by the end of the week. thanks again for your patience.

kevinoliver:
ok cool, i've gotten this patch merged. it'll show up on the `develop` branch real soon now. thanks again @jstultz

cacoco:
@jstultz this has been merged in dc119779ede0e80153068fff02642c9e0afcb038

PR #252, opened 2016-12-06
Revert "scrooge-generator: Handle exceptions thrown in decodeResponse"

cacoco:
Problem/Solution

Revert changes from RB 796974 while investigating a possible issue.

The RB ID listed is to ensure the internal->external syncing process
starts from the correct SHA.

RB_ID=894819
TBR=true


kevinoliver:
🚢 , thanks @cacoco

mosesn:
can we get in trouble for having discrepancies between the commits in source and in github?  how do we decide what to merge into github from source?

PR #258, opened 2017-02-08
Make Importer configurable in subclasses

johnynek:
Problem

While working to improve bazel support, I need to customize the importing to support relative imports (which our builds currently use). Currently we do this by extracting jars and doing modifications on the filesystem, but the jars are in some cases large and this is slow.

I want to customize the Importer, but I have no way to do that except by copy-paste of Compiler locally. If the importer was resolved in a method, I could subclass the Compiler.

Solution

I moved the creation of Importer to a protected method.

### DIFF
--- a/scrooge-generator/src/main/scala/com/twitter/scrooge/Compiler.scala
+++ b/scrooge-generator/src/main/scala/com/twitter/scrooge/Compiler.scala
@@ -47,6 +47,13 @@ class Compiler {
   var defaultNamespace: String = CompilerDefaults.defaultNamespace
   var scalaWarnOnJavaNSFallback: Boolean = false
 
+  /**
+   * Customizations may override this method to control
+   * the importer used
+   */
+  protected def getImporter: Importer =
+    Importer(new File(".")) +: Importer(includePaths)
+
   def run() {
     // if --gen-file-map is specified, prepare the map file.
     fileMapWriter = fileMapPath.map { path =>
@@ -61,7 +68,7 @@ class Compiler {
       new FileWriter(file)
     }
 
-    val importer = Importer(new File(".")) +: Importer(includePaths)
+    val importer = getImporter
 
     val isJava = language.equals("java")
     val documentCache = new TrieMap[String, Document]


ryanoneill:
Looks good to me @johnynek.

ryanoneill:
I will put this up for internal code review.

johnynek:
you can actually trash this.

You can see what I wound up doing here:
https://github.com/bazelbuild/rules_scala/pull/137

There were two problems:

1. I could not control the importer, but I needed more control than this PR, so I just copied the Compiler.scala code.
2. The ZipImporter does not work as I want (thrifts in a zip/jar cannot do relative imports). I fixed that with a custom Importer: https://github.com/bazelbuild/rules_scala/pull/137/files#diff-5dffef9fd92f934a1c5e238cc9faf0e9R18

This gave us a 65% win vs decompressing our jars before compiling.

Anyway, I am a little negative on sending a PR for two reasons:

1. I don't think you are actually going use this code.
2. your OSS build is pretty painful. I don't know if I want to invest the time to get this dodo thing working just to compile a PR. I hope you all will consider making sending PRs more friendly. There are a lot of folks at Stripe that would consider doing this, but as it is, it is somewhat compelling to fork so we can control generation ourselves (and we could still generate scrooge compatible classes, for instance).

kevinoliver:
> I don't know if I want to invest the time to get this dodo thing working just to compile a PR

@johnynek while its not as easy as `./sbt test`, is dodo actually hard to use?

From the [contributing docs](https://github.com/twitter/scrooge/blob/develop/CONTRIBUTING.md#building-dependencies) — you run `curl -s https://raw.githubusercontent.com/twitter/dodo/develop/bin/build | bash -s -- --no-test scrooge` to build its OSS dependencies. then its standard sbt development from there. and while sbt isn't my favorite build tool, it is standard.

johnynek:
dodo may or may not be hard to use, it may or may not work without any bugs that I need to diagnose, and that script may or may not pose a security risk to my work machine, but each of those unknowns pose a risk in time (and when you are N yaks-deps, you are trying to avoid getting to N+1). 

if this had just been `sbt test` I would have been happy to do it. Indeed sbt is the standard for scala projects, and has actually gotten quite good in the last 2 years in my opinion.

I get that your practice is better than not accepting patches at all, but in this case I would be working to give you code you don't want, as far as I know, nor would I benefit by you having (since me changing it then becomes harder).

PR #266, opened 2017-05-20
Map Enum type to TType.ENUM

simontoens:
Problem:

In generaed java thrift bindings, the enum type should map to TType.ENUM
instead of TType.I32.  I am wondering whether this is a bug or done on
purpose.

I found this issue while running the following scenario:

The thrift server is running code generated by a recent Scrooge version,
the client is running with code generated by an older Scrooge version (4.1).

The client calls a thrift method that has an enum argument, the enum type
is serialized and sent over the wire as TTYpe.ENUM.

The Scrooge-generated code on the server side looks like this:

public void read(TProtocol iprot) throws TException {
  TField field;
  iprot.readStructBegin();
  while (true)
  {
    field = iprot.readFieldBegin();
    if (field.type == TType.STOP) {
      break;
  }
  switch (field.id) {
    case 1:
    if (field.type == TType.I32) {    // <-----
      this.myType = MyType.findByValue(iprot.readI32());
    } else {
      TProtocolUtil.skip(iprot, field.type);
    }

The check field.type == TType.I32 is never true since field.type is a
TType.ENUM.

Solution:

Map Enum to the corresponding TType.

However, this change breaks backwards compatibility since the mapping of
enum -> I32 has been released already - in the example above, with this fix,
the client would still send a TType.I32 but the server would now expect a
TType.ENUM.  I guess this would have to be handled in the generated code.

### DIFF
--- a/scrooge-generator/src/main/scala/com/twitter/scrooge/java_generator/ApacheJavaGenerator.scala
+++ b/scrooge-generator/src/main/scala/com/twitter/scrooge/java_generator/ApacheJavaGenerator.scala
@@ -192,7 +192,7 @@ class ApacheJavaGenerator(
       case TI32 => "TType.I32"
       case TI64 => "TType.I64"
       case TDouble => "TType.DOUBLE"
-      case EnumType(enumValue, scope) => "TType.I32"
+      case EnumType(enumValue, scope) => "TType.ENUM"
       case StructType(structLike, scope) => "TType.STRUCT"
       case MapType(key, value, cpp) => "TType.MAP"
       case SetType(key, cpp) => "TType.SET"


mosesn:
@simontoens it looks like this was an old [bug][0] in thrift that was fixed 8 years ago.  we might just have never picked it up.  It looks like we read both enum and i32 permissively to support the migration, so I think this will be backwards compatible.  Could you test this out?

[0]: https://issues.apache.org/jira/browse/THRIFT-551

simontoens:
@mosesn thanks for the quick reply - yes, I can test this some more.

simontoens:
@mosesn I got back to this.  Looks like this change alone won't be enough - if a client without this change sends an enum value to a server that has this change, the value is dropped and the server gets null.  

Test setup:

Scrooge version: 4.16.0

thrift idl used for testing
```
namespace java mytest.thrift
enum TestEnum {
    ONE,
    TWO
}
service TestService {
    string test(1:TestEnum testEnum);
}
```

Server impl is:

```
public class TestServiceImpl implements TestService.ServiceIface {
    public Future<String> test(TestEnum testEnum) {
        System.out.println("Client sent a test enum: " + testEnum);
        return Future.value("test answer");
    }
```

Client:
```
        Thrift.Client client = Thrift.client();
        TestService.ServiceIface testClient = client.newIface("localhost:8090", TestService.ServiceIface.class);
        Future<String> testF = testClient.test(TestEnum.ONE);
        String result = Await.result(testF);
```

Without the change in this PR, this works as expected - output from server: "Client sent a test enum: ONE"

If the server then runs with code generated by Scrooge that has the change in this PR and the client remains unchanged, the output from the server becomes: "Client sent a test enum: null"

Relevant server-side generated code with the change in this PR:

```
      switch (field.id) {
        case 1: // TEST_ENUM
          if (field.type == TType.ENUM) {
            this.testEnum = TestEnum.findByValue(iprot.readI32());
          } else {
            TProtocolUtil.skip(iprot, field.type);
          }
          break;
        default:
          TProtocolUtil.skip(iprot, field.type);
      }
```

mosesn:
@simontoens OK, we should make sure the server can permissively read either I32 or Enum.  How does the other way work?  Client with the change talking to a server without the change?

simontoens:
@mosesn same behavior if the client has the change and the server doesn't because the client then sends the enum as TType.ENUM and the server expects TType.I32.  

So the fix here is probably that the read method needs to keep handling TType.I32 in addition to TType.ENUM

From my test:

Without change:

Generated code used by the client:
```
        public void write(TProtocol oprot) throws TException {
            this.validate();
            oprot.writeStructBegin(STRUCT_DESC);
            if(this.testEnum != null) {
                oprot.writeFieldBegin(TEST_ENUM_FIELD_DESC);
                oprot.writeI32(this.testEnum.getValue());
                oprot.writeFieldEnd();
            }

            oprot.writeFieldStop();
            oprot.writeStructEnd();
        }
```

The client sends TType.I32 as the type for TEST_ENUM_FIELD_DESC:
```
        private static final TField TEST_ENUM_FIELD_DESC = new TField("testEnum", TType.I32, 1);
```

With the change, the client sends TType.ENUM as the type for TEST_ENUM_FIELD_DESC:
```
        private static final TField TEST_ENUM_FIELD_DESC = new TField("testEnum", TType.ENUM, 1);
```

mosesn:
@simontoens OK, I think we're going to have to do this in phases then.  Immediately, we should upgrade servers to be able to read both TType.ENUM and TType.I32.  We should delay migrating clients to writing TType.ENUM to a subsequent commit though, since it will break old servers.  In the mean time, we can make it opt-in to write TType.ENUM instead of TType.I32 on the client-side.  Does that solve your problem?

simontoens:
@mosesn that sounds good thanks.  I am not immediately blocked by this because we have our own Scrooge fork.  If you want I can help with this work but I will be somewhat slow.

mosesn:
@simontoens would love your help on this!  We're ok with waiting =)

simontoens:
@mosesn sounds good I will close this for now then.

simontoens:
Hi @mosesn, I finally got back to this.  

Could you please have a look at this commit and let me know what you think?  If this is more or less what you were thinking, I will move on to adjusting the gold file tests.

https://github.com/simontoens/scrooge/commit/0b73a93e2d7d1f1a94bad7367c5e9c988efac82a

mosesn:
@simontoens yeah, that looks good to me!  does the scala generated client code already send enums, or should we update that too?  (if I remember correctly, the server already permissively parses enums)

simontoens:
@mosesn thanks for looking.  Yeah the generated Scala code seems fine:

```
              case 1 =>
                _field.`type` match {
                  case TType.I32 | TType.ENUM =>
                    testEnum = readTestEnumValue(_iprot)
```

I will send a real PR soon.

simontoens:
@mosesn nm, I realized you are asking about the client send, not the server reading.  I briefly looked at the Scala generation, as far as I can tell it is sending an I32 (TestEnumFieldI32 - and that behavior appears to be hardcoded):

```
      @inline private def writeTestEnumField(testEnum_item: mytest.thrift.TestEnum, _oprot: TProtocol): Unit = {
        _oprot.writeFieldBegin(TestEnumFieldI32)
        writeTestEnumValue(testEnum_item, _oprot)
        _oprot.writeFieldEnd()
      }
```
It would probably make sense to also make this behavior optional in Scala so that Java and Scala generated code behave the same.  However, I'd like to start with the Java-only PR.

mosesn:
@simontoens works for me!

PR #267, opened 2017-05-20
Removes @throws doc from Enum ethod

### LINKED ISSUE #220: Cannot build doc-jars with -Xfatal-warnings
We build our projects with the `-Xfatal-warnings` scalac option to enforce good hygiene.  Scrooge-generated enum types cause compilation failures, do to a scaldoc error:

```
[error] .../thriftscala/MyEnumKind.scala:34: Could not find any member to link for "NoSuchElementException".
[error]   /**
[error]   ^
```

The failing generated comment is:

```
/**
 * Find the enum by its integer value, as defined in the Thrift IDL.
 * @throws NoSuchElementException if the value is not found.
 */
```

I'm not exactly sure what the proper solution, but in the worst case it seems preferable to omit the @throws clause if it cannot build without warnings.

chrisbenincasa:
Fixes #220 

Problem

Because NoSuchElementException is defined in java.util
which is not automatically imported, scaladoc generation
generates warnings for generated enums. When
projects have fatal-warnings enabled, this fails builds.

Solution

Remove the offending scaladoc line

Result

scaladocs can be generated with fatal-warnings set

### DIFF
--- a/scrooge-generator-tests/src/test/resources/gold_file_output_scala/com/twitter/scrooge/test/gold/thriftscala/RequestType.scala
+++ b/scrooge-generator-tests/src/test/resources/gold_file_output_scala/com/twitter/scrooge/test/gold/thriftscala/RequestType.scala
@@ -40,7 +40,6 @@ case object RequestType {
 
   /**
    * Find the enum by its integer value, as defined in the Thrift IDL.
-   * @throws NoSuchElementException if the value is not found.
    */
   def apply(value: Int): com.twitter.scrooge.test.gold.thriftscala.RequestType =
     value match {
--- a/scrooge-generator/src/main/resources/scalagen/enum.scala
+++ b/scrooge-generator/src/main/resources/scalagen/enum.scala
@@ -34,7 +34,6 @@ case object {{EnumName}} {
 
   /**
    * Find the enum by its integer value, as defined in the Thrift IDL.
-   * @throws NoSuchElementException if the value is not found.
    */
   def apply(value: Int): {{package}}.{{EnumName}} =
     value match {


chrisbenincasa:
Is there any way to test this behavior? I tested manually by publishing a SNAPSHOT locally, including that artifact in a local repo, and generating docs.

mosesn:
@chrisbenincasa can you run the scrooge-generator-tests tests?  I think that it will probably fail the gold file tests.  Thanks for the PR!  How would you feel about calling it java.util.NoSuchElementException instead so that we can keep the line?

chrisbenincasa:
@mosesn I don't think that the problem is missing a full classpath, but rather that scaladoc cannot automatically link to external documentation (without an option) and even then, linking to Java classes is [more complicated](http://stackoverflow.com/questions/16934488/how-to-link-classes-from-jdk-into-scaladoc-generated-doc/31322970#31322970) because they'd live in javadoc. In this case, I think the only way to solve every case for downstream repos is to remove this line.

mosesn:
@chrisbenincasa I see, makes sense.  I'm restarting the travis build since it seems to have flaked in a way that doesn't make sense to me, but this LGTM.

chrisbenincasa:
@mosesn thanks! It's definitely unfortunate that there is no easy way to get this to work; I love rich documentation.

The error in the build is strange -- If i'm reading correctly, it's building finagle `develop` first, which requires a SNAPSHOT version of scrooge-sbt, but that's available yet (it wasn't published locally?)

mosesn:
@chrisbenincasa yeah, looking at the history, it looks like all PRs for months have been failing . . . it's something we should investigate on our end, but we have another separate CI infrastructure that we can  put it through to validate it, and the main travisci build is working, so let's not hold this up on travisci.  We'll pull it in Monday, thanks!

chrisbenincasa:
@mosesn cool, sounds great!

jcrossley:
@chrisbenincasa this has been merged internally and will go out with the next release. Thanks for your contribution!

PR #270, opened 2017-07-23
Fix deprecations in scrooge-sbt-plugin

3tty0n:
Problem

There are many deprecated methods in `ScroogeSBT.scala` like ( `<<=`, `<+=` ).

Solution

Change them to the recommended style.

Result

ScroogeSBTPlugin can be used in sbt 1.0.x.

### DIFF
--- a/scrooge-sbt-plugin/src/main/scala/com/twitter/ScroogeSBT.scala
+++ b/scrooge-sbt-plugin/src/main/scala/com/twitter/ScroogeSBT.scala
@@ -2,8 +2,7 @@ package com.twitter.scrooge
 
 import sbt._
 import Keys._
-
-import com.twitter.scrooge.backend.{WithFinagle, ServiceOption}
+import com.twitter.scrooge.backend.{ServiceOption, WithFinagle}
 import java.io.File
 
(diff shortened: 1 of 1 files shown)


mosesn:
@3tty0n I made a breaking API change to twitter commons, which makes it difficult for us to build scrooge & friends.  I'm planning on removing the dependency on twitter commons (which is why I made the change!) but in the mean time, we can't safely publish without risking breaking the build, so I'm going to hold off on merging this in for a few days until we remove the dependency completely.  Sorry for the delay!

mosesn:
@3tty0n finally got this out to github!  thanks for bearing with me.  f978e10e7c42c38784e6ca26f6f373f2bcee06c9

mosesn:
ugh wrong link.  your thing should be out in an hour or two, my bad!

mosesn:
Here's the right link, sorry about the confusion: https://github.com/twitter/scrooge/commit/5ebbf79a47871f8ac228fcb505462cdc2599dd8f

3tty0n:
@mosesn Thank you for your kind support!

PR #271, opened 2017-07-23
Migrate to build sbt

matsu-chara:
Problem:

Build trait was deprecated

Solution:

- step1: move project/Build.scala to build.sbt
- step2.1: remove Build trait
- step2.2: remove unnecessary import(sbt, Keys, and some plugins)
- step3: remove unnecessary settings
  - Defaults.coreDefaultSettings which is loaded by CorePlugin
  - JmhPlugin.projectSettings which is loaded by JmhPlugin

Result:
  Migration is completed


mosesn:
LGTM, thanks again!

mosesn:
@matsu-chara I made a breaking API change to twitter commons, which makes it difficult for us to build scrooge & friends.  I'm planning on removing the dependency on twitter commons (which is why I made the change!) but in the mean time, we can't safely publish without risking breaking the build, so I'm going to hold off on merging this in for a few days until we remove the dependency completely.  Sorry for the delay!

mosesn:
@matsu-chara OK, finally made it to github, thanks!  https://github.com/twitter/scrooge/commit/e7cb5875019eb56be5a45dbb210a13afe320a486

PR #272, opened 2017-07-26
Add new option: `--fatal-warnings`

3tty0n:
Problem:

There are many situations that we want to make warnings to errors.
For example, scala compiler has the option that converts warnings to errors (`-Xfatal-warnings`).
However, scrooge-linter has no function to do such a thing.

Solution:

We add a linter option that converts warnings to errors.

Result:

By adding `--fatal-warnings` , we can make warnings to errors.

### DIFF
--- a/scrooge-linter/src/main/scala/com/twitter/scrooge/linter/Linter.scala
+++ b/scrooge-linter/src/main/scala/com/twitter/scrooge/linter/Linter.scala
@@ -405,20 +405,33 @@ class Linter(cfg: Config) {
 
     val messages = LintRule.all(rules)(doc)
 
-    messages.foreach {
-      case LintMessage(msg, Error) =>
-        error(s"$inputFile\n$msg")
-      case LintMessage(msg, Warning) =>
-        warning(s"$inputFile\n$msg")
-    }
-
     val errorCount = messages.count(_.level == Error)
     val warnCount = messages.count(_.level == Warning)
 
-    if (errorCount + warnCount > 0 ) {
-      warning("%d warnings and %d errors found".format(messages.size - errorCount, errorCount))
+    if (cfg.fatalWarnings) {
+      val errorAndWarnCount = errorCount + warnCount
+      messages.foreach {
+        case LintMessage(msg, _) =>
+          error(s"$inputFile\n$msg")
+      }
+
+      if (errorAndWarnCount > 0) {
+        warning("%d warnings and %d errors found".format(0, errorAndWarnCount))
+      }
+      errorAndWarnCount
+    } else {
+      messages.foreach {
+        case LintMessage(msg, Error) =>
+          error(s"$inputFile\n$msg")
+        case LintMessage(msg, Warning) =>
+          warning(s"$inputFile\n$msg")
+      }
+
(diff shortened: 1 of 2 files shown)


scrooge-linter/src/main/scala/com/twitter/scrooge/linter/Linter.scala
```diff
@@ -405,20 +405,33 @@ class Linter(cfg: Config) {
...
+    } else {
+      messages.foreach {
+        case LintMessage(msg, Error) =>
+          error(s"$inputFile\n$msg")
+        case LintMessage(msg, Warning) =>
+          warning(s"$inputFile\n$msg")
+      }
+
+      if (errorAndWarnCount > 0) {
+        warning("%d warnings and %d errors found".format(messages.size - errorCount, errorCount))
```
jcrossley: We have warnCount above; let's use that instead of subtracting here.

scrooge-linter/src/main/scala/com/twitter/scrooge/linter/Linter.scala
```diff
@@ -405,20 +405,33 @@ class Linter(cfg: Config) {
...
-      case LintMessage(msg, Warning) =>
-        warning(s"$inputFile\n$msg")
-    }
-
     val errorCount = messages.count(_.level == Error)
     val warnCount = messages.count(_.level == Warning)
 
-    if (errorCount + warnCount > 0 ) {
-      warning("%d warnings and %d errors found".format(messages.size - errorCount, errorCount))
+    val errorAndWarnCount = errorCount + warnCount
```
jcrossley: let's move this down into the if(fatal warnings) block.

scrooge-linter/src/main/scala/com/twitter/scrooge/linter/Linter.scala
```diff
@@ -405,20 +405,33 @@ class Linter(cfg: Config) {
...
+
+    if (cfg.fatalWarnings) {
+      messages.foreach {
+        case LintMessage(msg, _) => error(s"$inputFile\n$msg")
```
jcrossley: mind moving this down a line so it looks like the other case matches?

scrooge-linter/src/main/scala/com/twitter/scrooge/linter/Linter.scala
3tty0n: fixed 👍

scrooge-linter/src/main/scala/com/twitter/scrooge/linter/Linter.scala
3tty0n: I think so. I moved it down.

scrooge-linter/src/main/scala/com/twitter/scrooge/linter/Linter.scala
3tty0n: Sure. I fixed it.

jcrossley:
This was merged internally and will go out with the next release, thank you!

3tty0n:
You're welcome! Thank you for your kind support!

PR #276, opened 2017-09-11
Add macro to the list of ScalaKeywords

avibryant:
The identifier `macro` is now a keyword in Scala. This adds it to the list of `ScalaKeywords` so that it gets escaped properly by Scrooge.


kevinoliver:
this got merged in https://github.com/twitter/scrooge/commit/3251125133d822994e1487738352148423e12014

thanks again, @avibryant

PR #282, opened 2017-11-12
Let scrooge-sbt-plugin use Scala 2.12

3tty0n:
Problem

Publishing scrooge-sbt-plugin is failed.

Solution

Update scalaVersion to 2.12.4, because sbt 1.0 requires sbt-plugins to use
Scala 2.12.

Result

We can publish scrooge-sbt-plugin now.


mosesn:
This works for me, but can you change the commit message a bit?  It's already possible today to publish scrooge-sbt-plugin, but you need to use `^^1.0.4` to indicate that you want to publish it for 1.0.x.  Instead, can we change the commit message to indicate that instead we're supporting 1.0.x as the default?

As a note for scrooge owners when we pull this patch in, we'll probably have to update the `dodo` script too.

3tty0n:
I will explain in more detail about that.

The command `sbt ";project scrooge-sbt-plugin;^^1.0.3 publishLocal"` is succeeded and I can publish scrooge-sbt-plugin to my local maven, but `sbt ";project scrooge-sbt-plugin; publishLocal"` is failed. 
And scala version in scrooge-sbt-plugin is still `2.10.6`, so I think this problem is caused by this.

I prefer the latter to former. Changing my commit message is OK :)

mosesn:
I see, that makes sense.  Can you update the commit message to explain that?

3tty0n:
@mosesn I just changed the commit message. Please take a look 🙏

build.sbt
```diff
@@ -293,9 +293,9 @@ lazy val scroogeSbtPlugin = Project(
 ).settings(
   // we remove the publishTo key because bintray already sets it, and setting
   // it again breaks the bintray plugin.
-  settingsWithTwoTen.filter(_.key.key.label != "publishTo")
+  sharedSettingsWithoutScalaVersion.filter(_.key.key.label != "publishTo")
```
mosesn: since we're changing this now, I don't think we ever use `settingsWithTwoTen` anymore.  can you delete that val?

mosesn:
Sorry, I missed your comment that you had updated the commit message.  Thanks!  This looks good, but it would be great to also clean up the val.

build.sbt
3tty0n: I removed this val in 1c690c2.

mosesn:
@3tty0n I think you misunderstood me.  In the previous commit, you changed which settings you were using from `settingsWithTwoTen` to `sharedSettingsWithoutScalaVersion`.  Once you had changed that, we never used `settingsWithTwoTen`, so I was thinking that we could just remove that field.  I'm pretty sure we still need to pass the settings to the project though.

mosesn:
@3tty0n sorry to see you close this ticket.  We'll try to fix it when we get some time.

PR #283, opened 2017-11-16
scrooge-linter: add lint rule to check type of map keys

DK777:
Problem

https://thrift.apache.org/docs/types
Thrift recommends that the key type of maps should be a basic type
as not all target languages/JSON support complex key types

Solution

I've added a linter rule to warn of complex key types.  I've kept it
a warning as many target languages do support complex key types.


yufangong:
Thanks @DK777 , this is merged in e8cd8ecbd569b7ed2d8f7c6554469a8eff8d348a.

DK777:
Thanks!

PR #291, opened 2018-04-04
scrooge-sbt-plugin: support sbt 1.1.x

3tty0n:
Problem

`scrooge-sbt-plugin` doesn't support for sbt 1.1.x. It seems that there are some people that want to use sbt 1.1.x, so supporting 1.1.x should be needed.

Solution

Enable cross build against 0.13.7 and 1.1.2. Please execute the command like `sbt ";project scrooge-sbt-plugin; ^ publish"`.

Result

We can use scrooge-sbt-plugin against 1.1.x.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -304,11 +304,12 @@ def scroogeSbtPluginSettings = {
 lazy val scroogeSbtPlugin = Project(
   id = "scrooge-sbt-plugin",
   base = file("scrooge-sbt-plugin")
-).enablePlugins(BuildInfoPlugin
+).enablePlugins(
+  BuildInfoPlugin
 ).settings(
   scroogeSbtPluginSettings: _*
 ).settings(
-  scalaVersion := "2.10.6",
+  crossSbtVersions := Vector("0.13.17", "1.1.2"),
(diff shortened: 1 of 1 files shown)


3tty0n:
@mosesn Hi, have you ever reviewed this PR?

mosesn:
@3tty0n OK, I have a patch ready to go for our internal script, but I'd like to confirm with you that 1.1.x seems to work for you before we merge it in.  Unfortunately, we can't test it ourselves yet because we've been running into issues with 1.1.x sbt.

3tty0n:
@mosesn Sorry for my late reply because I've been busy lately. I'm going to test to work with sbt 1.1.x in our internal system. Could you please wait for a while?

mosesn:
yep, that works for me!  we've actually been able to get 1.1.x going on our system too, so we might be able to test it ourselves too.

3tty0n:
I ran into some trivial problems, when I tested in our internal system with sbt 1.1.x. But I cannot figure out why it was happened. Next, I'm going to investigate the cause of this problem.

mosesn:
@3tty0n it turns out that there isn't a separate 1.1 namespace, so you should be able to use our 1.0 namespaced stuff with sbt 1.1.  https://github.com/sbt/sbt/issues/3858

PR #300, opened 2019-01-19
Add Java 11 compatibility.

### LINKED ISSUE #299: JDK 9+ compilation failed with wrong annotation "javax.annotation.Generated"
JDK 9+ not supported yet.

### Expected behavior

we can compile `.thrift` files under JDK 9, JDK 10 and JDK 11.

### Actual behavior

got compilation errors:

```
[error] ... type Generated is not a member of package javax.annotation
[error] @javax.annotation.Generated(value = Array("com.twitter.scrooge.Compiler"))
```

- Java 9: javax.annotation.processing.Generated
- Java 8: javax.annotation.Generated

### Steps to reproduce the behavior

run `sbt compile` to compile `.thrift` files under JDK 11.

mmlac:
Fixes #299

- Added Java 11 compliant `Generated` annotation when running on Java 11
- Updated tests for Java 8 as well as Java 11
  - Tests will compare to Java 8 gold files when running on Java 8
  - Tests will compare to Java 11 gold files when running on Java 11
  - Tests will not run on Java 9 or 10, as those have already been
  superseded by Java 11

Caveat: I was unable to build scrooge-sbt-plugin with 2.10.6 because of what I assume to be compiler version issues (see below) but I assume these are issues with my environment and not with the code itself. After moving the version of scrooge-sbt-plugin to 2.12.7 (the current version of SBT) the build passed and the plugin worked as expected on Java 11 on my test project.

```
[error]   last tree to typer: Literal(Constant(thrift))
[error]               symbol: null
[error]    symbol definition: null
[error]                  tpe: String("thrift")
[error]        symbol owners: 
[error]       context owners: value ThriftConfig -> object ScroogeSBT -> package scrooge
[error] 
[error] == Enclosing template or block ==
[error] 
[error] Import( // val <import>: ImportType(ScroogeSBT.this.autoImport)
[error]   "autoImport" // object autoImport in object ScroogeSBT, tree.tpe=com.twitter.scrooge.ScroogeSBT.autoImport.type
[error]   List(
[error]     ImportSelector(_,5139,null,-1)
[error]   )
[error] )
[error] 
[error] == Expanded type of tree ==
[error] 
[error] ConstantType(value = Constant(thrift))
[error] 
[error] uncaught exception during compilation: scala.MatchError
[error] ## Exception when compiling 2 sources to /home/xyz/scrooge/scrooge-sbt-plugin/target/scala-2.10/sbt-1.0/classes
[error] false (of class scala.reflect.internal.Trees$Literal)
[error] scala.tools.nsc.typechecker.Macros$MacroImplBinding$.unpickleAtom(Macros.scala:118)
[error] scala.tools.nsc.typechecker.Macros$MacroImplBinding$$anonfun$7.apply(Macros.scala:186)
[error] scala.tools.nsc.typechecker.Macros$MacroImplBinding$$anonfun$7.apply(Macros.scala:186)
[error] scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:244)
[error] scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:244)
[error] scala.collection.immutable.List.foreach(List.scala:318)
[error] scala.collection.TraversableLike$class.map(TraversableLike.scala:244)
[error] scala.collection.AbstractTraversable.map(Traversable.scala:105)
[error] scala.tools.nsc.typechecker.Macros$MacroImplBinding$.unpickle(Macros.scala:186)
```


mosesn:
@mmlac oh geez, sorry that we ignored this for so long.  are you still interested in this work, or are you busy now?

mosesn:
From talking to someone on my team who has been working on jdk11 compat, she said that she has had success with running scrooge in its current form on jdk11.  Is your concern around compile-time issues?

mmlac:
Hey @mosesn, the issue isn't running scrooge under JRE 11, the issue is that it will generate code that is invalid to run under Java 11. So if you have a Java 11 project and Scrooge generates the code, that code doesn't compile because the annotation moved between Java 8 and 11. This simple fix addresses that and generates the correct annotations. Unless you fixed it in master already — I haven't checked recently — this PR makes the _output_ code Java 11 compatible.

mosesn:
@mmlac got it, yeah, I think we're seeing the problem you're describing now.  This is really useful, I think we're probably going to pull this in.  Would you mind merging master for us?

scrooge-generator/src/main/scala/com/twitter/scrooge/Util.scala
```diff
@@ -0,0 +1,18 @@
+package com.twitter.scrooge
+
+/**
```
mosesn: here and elsewhere, we use the other style of comments:

```
/**
 *
 */
```

mosesn:
@mmlac do you think it might make sense to add a dependency on javax.annotation:javax.annotation-api in jdk11 instead of moving forward with this approach?  that way things that are compiled for jdk8 can also be used in jdk11

cacoco:
As a start the `javax.annotation` dependency was added here: https://github.com/twitter/scrooge/commit/78b06e92b9a5268f1c684eb186f7bff1dd429886.

We're working on JDK 11 support in all of our libraries. Hopeful to have this addressed soon, thanks!

PR #310, opened 2019-06-11
scrooge-generator: create ThriftMethodStats once

jstultz:
Problem

The current implementation of per endpoint stats creates a newly scoped
stats receiver and instance of ThriftMethodStats on each request; it
consequently also creates counters for requests/success/failures on each
request. For some StatsReceivers, this may be expensive, or at least,
not trivial.

Solution

This moves the creation of ThriftMethodStats into perMethodStatsFilter,
where a single instance is (optionally) created for the filter, and then
passed into updated recordRequest/recordResponse methods.

In addition to being more efficient, this also deduplicates code a
little bit.

It's worth noting that when perEndpointStats is false, it appears that
both recordRequest and recordResponse are no-ops; I considered
reorganizing the code to only call them if perEndpointStats is true
(e.g. methodStats.foreach(recordRequest) rather than
 recordRequest(methodStats)) but opted to keep the scope of the change
smaller for now.

### DIFF
--- a/scrooge-generator-tests/src/test/resources/gold_file_output_scala/com/twitter/scrooge/test/gold/thriftscala/GoldService.scala
+++ b/scrooge-generator-tests/src/test/resources/gold_file_output_scala/com/twitter/scrooge/test/gold/thriftscala/GoldService.scala
@@ -955,6 +955,8 @@ object GoldService extends _root_.com.twitter.finagle.thrift.GeneratedThriftServ
           stats.counter("failures"),
           stats.scope("failures")
         )
+
+      val nullMethodStats = apply(com.twitter.finagle.stats.NullStatsReceiver)
     }
 
     private case class ThriftMethodStats(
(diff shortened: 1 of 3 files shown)


scrooge-generator-tests/src/test/resources/gold_file_output_scala/com/twitter/scrooge/test/gold/thriftscala/GoldService.scala
```diff
@@ -982,27 +982,25 @@ object GoldService extends _root_.com.twitter.finagle.thrift.GeneratedThriftServ
         }
       }
 
-    private def recordRequest(method: ThriftMethod): Unit = {
-      if (perEndpointStats) {
-        val methodStats = ThriftMethodStats((if (serviceName != "") stats.scope(serviceName) else stats).scope(method.name))
-        methodStats.requestsCounter.incr()
+    private def recordRequest(methodStats: Option[ThriftMethodStats]): Unit = {
+      methodStats.foreach { stats =>
```
kevinoliver: we like to avoid `Option.foreach` as it creates a `Function1`. instead could you use:
```
methodStats match {
  case Some(stats) => ...
  case None => ()
}
```

scrooge-generator-tests/src/test/resources/gold_file_output_scala/com/twitter/scrooge/test/gold/thriftscala/PlatinumService.scala
```diff
@@ -1083,27 +1083,25 @@ object PlatinumService extends _root_.com.twitter.finagle.thrift.GeneratedThrift
...
+        stats.requestsCounter.incr()
       }
     }
 
-    private def recordResponse(reqRep: ctfs.ReqRep, method: ThriftMethod): Unit = {
+    private def recordResponse(reqRep: ctfs.ReqRep, methodStats: Option[ThriftMethodStats]): Unit = {
       ServerToReqRep.setCtx(reqRep)
-      if (perEndpointStats) {
-        val methodStats = ThriftMethodStats((if (serviceName != "") stats.scope(serviceName) else stats).scope(method.name))
+      methodStats.foreach { stats =>
```
kevinoliver: similar comment here around avoiding the `Option.foreach`. 

although, actually. here's another approach that i think i would prefer.

instead of using None to represent that case, create a singleton instance of `ThriftMethodStats` that uses a `NullStatsReceiver`. this instance could live on the ThriftMethodStats companion object. or create one as needed, either way.

jstultz: That sounds better to me as well. Before I go ahead and make that change, though, do you think it might make more sense to just bypass calls to these methods entirely in the event that perEndpointStats is false?

kevinoliver: my preference is the simpler code. that's why i like the "sentinel" object approach leaving just one code path.

jstultz:
Alright, updated to use a single instance of ThriftMethodStats defined on the companion object with the NullStatsReceiver rather than using an Option.

mosesn:
thanks @jstultz this is rad!!

mosesn:
this is great, thanks @jstultz!

kevinoliver:
Thanks again. This got merged in ddf6c924b668c8c2091425b9c4b77f65c0bedd8a.

Issue #308, opened 2019-05-10
Support Scala 2.13 RC1 and beyond


skonto:
`scrooge-serializer` is required by several projects, like chill which is used again by many other projects like Spark. Would be good to start supporting 2.13 as it will be soon out.

ayoub-benali:
tried to build but it seems to be blocked by this PR https://github.com/twitter/util/pull/248 at least

martijnhoekstra:
twitter/util is released, but not yet published.

As long as sbt is stuck on 1.1.4 for cross-publishing, we can't use sbt to build.

There is a draft for 2.13 support at https://github.com/twitter/scrooge/compare/develop...martijnhoekstra:213?expand=1  but I don't really understand the code so I'm not entirely sure it's correct, even though it's a minimal diff.

In addition, it relies on sbt 2.13.0-RC3 there to (hopefully?) avoid the sbt 1.2.8 issue. A stable sbt version would probably be much preferable. 1.1.4 doesn't work with scala 2.13

ayoub-benali:
> it relies on sbt 2.13.0-RC3 there to (hopefully?) 

You mean sbt `1.3.0-RC3` ?

I don't see a problem kicking off a PR with sbt `1.3.0-RC3` by the time we are about to merge it we either upgrade sbt or make an other PR later.

So far we are still blocked by https://github.com/twitter/util/issues/253

martijnhoekstra:
Right, `1.3.0-RC3`

Is it possible to coerce both in at the same time? It would be a shame if we had to wait for a 19.9 release of twitter util, and then miss the 19.9 release for scrooge, and have to bring it along with 19.10 if my branch is indeed sufficient.

ayoub-benali:
> Is it possible to coerce both in at the same time

Unless there is a valid reason to stick to sbt `1.1.4` we should push for an update. 

The release schedule you describe is very bad :/ 
Do all these lib need to be released at the exact same time ? Otherwise we might be able to release them all in the same month just with few hours/days of difference

martijnhoekstra:
Yeah, that's what I was asking about: is it possible to get both scrooge and util updated in the same release cycle.

ayoub-benali:
> Yeah, that's what I was asking about: is it possible to get both scrooge and util updated in the same release cycle.

@yufangong can you or other maintainers help with this ?

SethTisue:
twitter-util has now been published for 2.13

ayoub-benali:
[...]

PR #314, opened 2019-08-15
cross-build for 2.13

martijnhoekstra:
Problem

No 2.13 build

Solution

Add one


martijnhoekstra:
fixes https://github.com/twitter/scrooge/issues/308
blocked by https://github.com/twitter/util/issues/253

martijnhoekstra:
This is ready for review. Not everything builds on 2.13 yet; there are circular dependencies between scrooge-generator-tests and finagle. ping @cacoco

martijnhoekstra:
My approach here was fuzzing the code until it worked -- I had no idea what I was doing.

`$` in identifiers have been reserved for compiler internal use for quite a while (which probably means having it in user code is considered a bug), and migrating away is probably a good idea -- from the looks of things `$` never leaks out to a public interface, so it shouldn't break anything. I don't know if it will actually start emitting warnings or errors at some point.

I'm not qualified/available for that job though, as it will probably need to be done by someone who actually understands what the code is doing.

I could s/\$//g and see if the tests pass though.

mosesn:
oh, I see, the immutable$Map is a rename.  since it seems to be working let's keep it this way for now, and we'll clean it up later.  I have an internal ticket to clean up other $-related stuff in scrooge anyway.

PR #317, opened 2019-09-26
scrooge-linter: warn when function names are reserved words

gjfishman:
Problem

The Apache Thrift compiler will fail to compile if it encounters struct, field, or function names that contain reserved words from various programming languages. However, Scrooge will compile these files successfully. This can cause problems when one developer creates a service using Scrooge, and another developer tries to create a client for this service using Apache Thrift.

Solution

- Expanded the linter "Keywords" rule to include function names in addition to Struct and field names
- Added Javascript and Go keywords to the lists of keywords to warn on.

Result

Users will get a warning from the linter if they attempt to use a keyword as a function name.
Users will get a warning from the linter if they attempt to use a Javascript or Go keyword as a struct, field, or function name.

Problem

Explain the context and why you're making that change.  What is the
problem you're trying to solve? In some cases there is not a problem
and this can be thought of being the motivation for your change.

Solution

Describe the modifications you've done.

Result

What will change as a result of your pull request? Note that sometimes
this section is unnecessary because it is self-explanatory based on
the solution.

### DIFF
--- a/doc/src/sphinx/Linter.rst
+++ b/doc/src/sphinx/Linter.rst
@@ -10,9 +10,9 @@ To run the thrift linter with sbt:
 
 .. code-block:: bash
 
-    $ ./sbt 'scrooge-linter/run-main com.twitter.scrooge.linter.Main --help'
+    $ ./sbt 'scrooge-linter/runMain com.twitter.scrooge.linter.Main --help'
 
-    $ ./sbt 'scrooge-linter/run-main com.twitter.scrooge.linter.Main /path/to/thrift/file.thrift'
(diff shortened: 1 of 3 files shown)


mosesn:
Does Thrift have a mega-list of reserved keywords it shares across languages, or is it language-specific?  If there's a mega-list, can you link us to that list?

gjfishman:
Thrift has one big list for all languages: https://github.com/apache/thrift/blob/master/compiler/cpp/src/thrift/generate/t_generator.cc#L80

cacoco:
Merged in ca2ae718ca224434631e8b2809242e6bb5788ffa

Issue #44, opened 2013-02-08
Fully Qualified Type Names are not supported in struct/service defs


hbs:
We have a best practice of using fully qualified names when referencing a Thrift struct in another Thrift IDL, this practice is not supported by the Scrooge generator which therefore is not backward compatible with the standard Thrift compiler.

The standard Thrift compiler accepts the file name (minus the '.thrift' suffix) as a prefix for the structures used in another file, Scrooge doesn't :-( 

file: foo.bar.thrift.data.thrift

namespace java foo.bar.thrift.data

struct A {
  1: bool bool_field,
}

file: foo.var.thrift.services.thrift

namespace java foo.bar.thrift.services

include "foo.bar.thrift.data.thrift"

service Svc {
  list &lt; foo.bar.thrift.data.A &gt; get(1:string name)
}

Calling thrift --gen java:beans foo.bar.thrift.services.thrift correctly generates code when
scrooge -l java foo.bar.thrift.services.thrift (or foo.bar.thrift.*.thrift) produces the following error:

Exception in thread "main" com.twitter.scrooge.TypeNotFoundException: foo.bar.thrift.data.A
	at com.twitter.scrooge.TypeResolver$$anon$1.createMissingEntityException(TypeResolver.scala:74)
	at com.twitter.scrooge.EntityResolver.apply(TypeResolver.scala:60)
	at com.twitter.scrooge.TypeResolver.apply(TypeResolver.scala:274)
	at com.twitter.scrooge.TypeResolver.apply(TypeResolver.scala:201)
	at com.twitter.scrooge.TypeResolver.apply(TypeResolver.scala:204)
	at com.twitter.scrooge.TypeResolver.apply(TypeResolver.scala:197)
	at com.twitter.scrooge.TypeResolver.apply(TypeResolver.scala:185)
	at com.twitter.scrooge.TypeResolver$$anonfun$7.apply(TypeResolver.scala:171)
	at com.twitter.scrooge.TypeResolver$$anonfun$7.apply(TypeResolver.scala:171)
	at scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:233)
	at scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:233)
	at scala.collection.LinearSeqOptimized$class.foreach(LinearSeqOptimized.scala:59)
	at scala.collection.immutable.List.foreach(List.scala:76)
[...]

PR #318, opened 2020-01-29
Various fixes for scrooge-generator

chenzhuoyu:
# Problems

1. Enumeration class name conflicts with class members.

Consider the following IDL:

```thrift
enum OrganizationType {
    TYPE_A = 1
    TYPE_B = 2
}

struct Organization {
    1: OrganizationType OrganizationType
}
```

This will be converted to something like:

```java
{
    TList _list430 = iprot.readListBegin();
    this.OrganizationType = new ArrayList<OrganizationType>(_list430.size);
    for (int _i431 = 0; _i431 < _list430.size; ++_i431)
    {
        OrganizationType _elem432;
        _elem432 = OrganizationType.findByValue(iprot.readI32());
        //         ^--------------- this conflicts with the field name
        this.OrganizationType.add(_elem432);
    }
    iprot.readListEnd();
}
```

Which cannot compile.

2. Scrooge generates invalid string escape sequences when the field annotation contains certain characters.

For example:

```thrift
struct Foo {
    1: i32 bar (tag1 = "\"value\"")
}
```

This will be converted to something like:

```java
{
    Map<String, String> tmpFieldMap = new HashMap<String, String>();
    tmpFieldMap.put("tag1", "\&quote;value\&quote;");
    //                       ^------- invalid escape sequences
    tmpMap.put(_Fields.TENANT_ID, Collections.unmodifiableMap(tmpFieldMap));
}
```

3. Currently scrooge-generator does not support more than one dot in thrift file name.

This might be related to issue #44 , but since it's been 7 years, I decided to fix it myself.

Suppose we have a thrift file with the name `com.whatever.product.module1.thrift`:

```thrift
struct Foo {
    1: i32 foo_field;
}
```

... and another thrift file that includes `com.whatever.product.module1.thrift`: 

```thrift
include "com.whatever.product.module1.thrift"
struct Bar {
    1: com.whatever.product.module1.Foo bar_field;
}
```

This doesn't work with the current version of scrooge-generator.

# Solutions

1. By replacing the type name with a fully-qualified one solves the problem.
2. Mustache will html-escape everything that is wrapped by `{{...}}`. By replacing the `{{value}}` with `{{{value}}}` solves the problem.
3. I found in the code what appears to be assuming the base part of a file name to be a `SimpleID`, which in fact isn't the case. Changing the `scopePrefix` to `Identifier` & other related places fixes this problem.

### DIFF
--- a/scrooge-generator/src/main/resources/androidgen/generate_deserialize_field.mustache
+++ b/scrooge-generator/src/main/resources/androidgen/generate_deserialize_field.mustache
@@ -40,6 +40,6 @@ iprot.read{{field_type.get_type}}End();
 {{/field_type.is_base_type_or_binary}}
 
 {{#field_type.is_enum}}
-{{name}} = {{{field_type.type_name_in_container}}}.findByValue(iprot.readI32());
+{{name}} = {{{field_type.qualified_type_name}}}.findByValue(iprot.readI32());
 {{/field_type.is_enum}}
 {{/consolidate_newlines}}
--- a/scrooge-generator/src/main/resources/apachejavagen/generate_deserialize_field.mustache
+++ b/scrooge-generator/src/main/resources/apachejavagen/generate_deserialize_field.mustache
@@ -49,6 +49,6 @@ iprot.read{{field_type.get_type}}End();
 {{/field_type.is_base_type_or_binary}}
 
 {{#field_type.is_enum}}
-{{name}} = {{{field_type.type_name_in_container}}}.findByValue(iprot.readI32());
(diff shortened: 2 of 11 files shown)


chenzhuoyu:
Sorry I pressed the wrong key before I finished my PR description ....

dotordogh:
Thanks for the submission! Could you please split this PR into 3 separate PRs that we can take a look at?

chenzhuoyu:
Sure ~

chenzhuoyu:
@dotordogh I've split this PR in to 3:
https://github.com/twitter/scrooge/pull/319
https://github.com/twitter/scrooge/pull/320
https://github.com/twitter/scrooge/pull/321

PR #320, opened 2020-02-01
fix: invalid string escape in annotations

chenzhuoyu:
# Problem

Scrooge generates invalid string escape sequences when the field annotation contains certain characters.

For example:

```thrift
struct Foo {
    1: i32 bar (tag1 = "\"value\"")
}
```

This will be converted to something like:

```java
{
    Map<String, String> tmpFieldMap = new HashMap<String, String>();
    tmpFieldMap.put("tag1", "\&quote;value\&quote;");
    //                       ^------- invalid escape sequences
    tmpMap.put(_Fields.TENANT_ID, Collections.unmodifiableMap(tmpFieldMap));
}
```

# Solution

Mustache will html-escape everything that is wrapped by `{{...}}`. By replacing the `{{value}}` with `{{{value}}}` solves the problem.

### DIFF
--- a/scrooge-generator-tests/src/test/resources/test_thrift/annotations.thrift
+++ b/scrooge-generator-tests/src/test/resources/test_thrift/annotations.thrift
@@ -1,0 +1,5 @@
+namespace java test.annotations
+
+struct FooStruct {
+    1: i32 field (tag1 = "\"value\"")
+}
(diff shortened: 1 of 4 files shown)


dotordogh:
Thank you @chenzhuoyu! Please also add tests for this as well to ensure consistent behavior.

mosesn:
@chenzhuoyu have you had a chance to take a stab at adding a test case?

chenzhuoyu:
@mosesn @dotordogh Test case added.

bryce-anderson:
@chenzhuoyu, another one that requires a rebase and squash.

chenzhuoyu:
@bryce-anderson Conflicts resolved.

yufangong:
Hi @chenzhuoyu, thanks for your PR, we are getting it merged. We would like to hear more context about why you need this supported (triple stashes). Thanks!

yufangong:
merged at https://github.com/twitter/scrooge/commit/a67086d34465ca4f980f2ad170ce5418ee94df03

PR #333, opened 2020-07-31
avoid deprecated ScalaTest 3.0 APIs

SethTisue:
Problem                                                                                                                 
                                                                                                                        
The codebase was still using deprecated ScalaTest 3.0 APIs. This was problematic in the Scala 2.13 community build, and it also                                                          
blocks an eventual upgrade to ScalaTest 3.2, where the deprecated APIs are removed.                                                                                                       
                                                                                                                        
Solution                                                                                                                
                                                                                                                        
Use the replacement 3.1/3.2 APIs recommended by the maintainers of ScalaTest.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -61,7 +61,7 @@ the maintainers will help out during code review.
 We've standardized on using the [ScalaTest testing framework][scalatest].
 Because ScalaTest has such a big surface area, we use a restricted subset of it
 in our tests to keep them easy to read.  We've chosen the `assert` API, not the
-`Matchers` one, and we use the [`FunSuite` mixin][funsuite], which supports
+`Matchers` one, and we use the [`AnyFunSuite` mixin][funsuite], which supports
 xUnit-like semantics.
 
 Note that while you will see a [Travis CI][travis-ci] status message in your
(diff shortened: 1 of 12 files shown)


cacoco:
@SethTisue thanks for the PR, we're in the midst of a rolling update of these deprecations internally but we should be able to merge this in. Thanks again!

mosesn (approved):
thank you!!

jyanJing:
Hi @SethTisue , thank you for making the change!
Could you change the description of the issue to include `Problem`, `Solution`, `Result` sections, then I can merge this? Example:

`Problem

Explain the context, and why you're making the respective change.
What is the problem you're trying to solve? This is usually higher level and omits any details about how to solve the problem or the details of the diff.
 
Solution

Describe the modifications you've done (i.e. the mechanics of the diff).
 
Result
 
After your change, what will change. If the result is obvious, this can be omitted.`

Thank you!

jyanJing (approved):

SethTisue:
done

jyanJing:
Merged in [9878247](https://github.com/twitter/scrooge/commit/9878247b1bd474fcbb03e3044b664e289565f652)

PR #340, opened 2020-11-22
Using irrefutable extractors for Scala structs

ncreep:
## Problem

Up until [Scala 2.13.4](https://github.com/scala/scala/pull/9140), having custom extractors in a pattern match disables exhaustivity  checks by the compiler. 
This is true, unless the extractor is "irrefutable", which means that its type-signature contains a `Some` rather than an `Option`.
For more details, see [this](https://nrinaudo.github.io/scala-best-practices/unsafe/custom_extractors.html).

As currently implemented, Thrift structs generate Scala code with "refutable" custom extractors, which can lead to inadvertent and surprising cases where people expect the compiler to do full exhaustivity checking, but that doesn't happen, leading to runtime exceptions.

## Solution

This pull request updates the relevant mustache template to have `Some` in the `unapply` signature, thus turning the extractor into an irrefutable one.

(Note, I'm not quite sure how to go about tests in this project, any pointers would be welcome.)

Thanks

### DIFF
--- a/scrooge-generator-tests/src/test/resources/gold_file_output_scala/com/twitter/scrooge/test/gold/thriftscala/AnotherException.scala
+++ b/scrooge-generator-tests/src/test/resources/gold_file_output_scala/com/twitter/scrooge/test/gold/thriftscala/AnotherException.scala
@@ -155,7 +155,7 @@ object AnotherException extends ValidatingThriftStructCodec3[AnotherException] w
       errorCode
     )
 
-  def unapply(_item: AnotherException): _root_.scala.Option[Int] = _root_.scala.Some(_item.errorCode)
+  def unapply(_item: AnotherException): _root_.scala.Some[Int] = _root_.scala.Some(_item.errorCode)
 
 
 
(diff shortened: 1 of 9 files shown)


ryanoneill:
Hi @ncreep. I ran this through an initial pass internally, and there are some significant challenges for us (Twitter) to work through in our own code for this to ship. I'm not sure yet what that will fully entail, but I don't see this as something which will be quick to resolve. We will keep you posted as we learn more.

ncreep:
Thanks for looking into this, let me know if there's anything I can do to help.

ryanoneill:
On first pass, this causes tens of thousands of failures internally. I'm seeing if I can knock that number down to something reasonable quite quickly. If not, we will be unable to ship this change for now. We should know more sometime early next week.

ncreep:
Wow...

Out of curiosity, what sort of failures does this cause?
I would've imagined that it will only trigger pattern-match exhaustiveness warnings and maybe some type-inference issues (unless reflection is somehow involved).

ryanoneill:
For us, it causes pattern-match exhaustiveness failures. In a few cases that I've seen so far, I've seen code that handles `Return(xyz(something))` and `Throw(exc)` and that logically covers all scenarios, yet the exhaustivity checker will complain that that code isn't handling `Return(_)` which is _right_, but not going to happen in that particular case. 

One instance of that in a shared library used by many leaf nodes can lead to ~5000 target failures. If that's the case, and it's really less than a few hundred files causing problems, then we maybe can ship this. If it's really thousands of instances of this occurring though, it's likely not something we can do at the moment.

ncreep:
Ah, that makes sense.

Too bad if this won't go in. But seeing how you probably invested more time into this then I spent on my changes, I appreciate the effort.

PR #354, opened 2022-07-27
Test out if this helps

vkostyukov:
This may help. This may not help. This is SBT after all, right?


cacoco:
@vkostyukov note you likely want to bust the GitHub action cache. @dotordogh knows how.

